PR #18, opened 2024-06-05
Degree-based orderings

gdalle:
- [x] Implement 3 dynamic degree-based orders following the ColPack paper: `SmallestLast`, `IncidenceDegree`, `DynamicLargestFirst`
- [ ] Generalize them to `BipartiteGraph` in addition to `Graph`
- [ ] Test that they are correct


amontoison:
Should we add it in the release 0.4.0 ?

gdalle:
It's not ready yet and it's not breaking so I'd say no. Bringing it up to speed with the latest developments isn't much work, the trouble is rather finding ways to test that the orders are correct

PR #26, opened 2024-07-17
Add decompress_row! and decompress_column!

amontoison:
- Add `decompress_row!` and `decompress_column!`
- Fix a typo in the docstring of `decompress_rows!` and `decompress_rows`

cc @gdalle


gdalle:
I understand the purpose of this, but for now let's keep it on the back burner. It is conceptually simpler to reason in terms of full matrices, and when we move to smaller colorings that require linear systems, we'll need the full matrix anyway

gdalle:
This no longer fits within the new framework, closing while we figure out if we really need it (note that it wouldn't work with substitution decompressionà

PR #27, opened 2024-07-17
Optimize decompress_symmetric!

amontoison:
@gdalle 
- Exploit the symmetry in `decompress_symmetric!`.

### DIFF
--- a/src/decompression.jl
+++ b/src/decompression.jl
@@ -188,6 +188,7 @@ function decompress_symmetric!(
     group = color_groups(color)
     for ij in findall(!iszero, S)
         i, j = Tuple(ij)
+        (i > j) && continue
         j2_exists = false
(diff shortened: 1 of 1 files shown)


gdalle:
I had thought about that but DI has no convention to return either the top or bottom triangle of the Hessian, so to be safe I compute both.

src/decompression.jl
```diff
@@ -223,7 +224,7 @@ function decompress_symmetric!(
         end
     end
     for ((i, j), star_id) in pairs(star)
-        i == j && continue
+        (i ≥ j) && continue
```
amontoison, line 227:
```suggestion
        (i == j) && continue
```

gdalle:
Superseded by #29, which ensures that each nonzero is visited exactly once

Issue #35, opened 2024-08-02
Should decompression be represented as a matrix multiplication?


gdalle:
It is a linear operation so yes. However, it is a linear operation on vectors of values, not matrices.

For column decompression, let $A \in \mathbb{R}^{m \times n}$ be the original matrix and $B \in \mathbb{R}^{m \times c}$ be the compressed one.
Then there exists a matrix $C \in \mathbb{R}^{mn \times mc}$ such that $\mathrm{vec}(A) = C \mathrm{vec}(B)$.
We can further replace $\mathrm{vec}(A)$ with $\mathrm{nonzeros(A)}$ and then $C$ is block-diagonal with exactly one $1$ per row and at most one $1$ per column.

Example: if

$$
A = \begin{pmatrix}
a_{11} & 0 & 0 & a_{14} \\
0 & a_{22} & 0 & a_{24} \\
a_{31} & 0 & a_{33} & 0 \\
a_{41} & 0 & 0 & 0
\end{pmatrix}
\quad \text{and} \quad
B = \begin{pmatrix}
a_{11} & a_{14} \\
a_{22} & a_{24} \\
a_{31} & a_{33} \\
a_{41} & 0
\end{pmatrix}
$$

then we have

$$
\begin{pmatrix}
a_{11} \\
a_{31} \\
a_{41} \\
a_{22} \\
a_{14} \\
a_{24} \\
a_{33}
\end{pmatrix}
= \begin{pmatrix}
1 & . & . & . & . & . & . & . \\
. & . & 1 & . & . & . & . & . \\
. & . & . & 1 & . & . & . & . \\
. & 1 & . & . & . & . & . & . \\
. & . & . & . & . & . & 1 & . \\
. & . & . & . & 1 & . & . & . \\
. & . & . & . & . & 1 & . & . \\
\end{pmatrix}
\begin{pmatrix}
b_{11} \\
b_{12} \\
b_{13} \\
b_{14} \\
b_{21} \\
b_{22} \\
b_{23} \\
b_{24}
\end{pmatrix}
$$

Is this useful in practice? How does it generalize
- to symmetric matrices? 
- to substitution methods where a linear system must be solved? 
- to matrices stored as non-sparse formats (e.g. tridiagonal)?

In all cases we have linear functions.

PR #36, opened 2024-08-05
Add an acyclic coloring

amontoison:
_modified by @gdalle_

- [x] Implement `acyclic_coloring` with the efficient tree-based algorithm (adds DataStructures.jl to the deps).
- [x] Implement very naive decompression by substitution for symmetric matrices, can be optimized much more even without the `TreeSet`. It works by building a matrix `T` such that `T * upper_nonzeros(A) == vec(B)` and then inverting the system.
- [x] Add tests on small instances from the papers as well as large random instances (the equality test is only approximate).


gdalle:
Thanks for contributing this! However, before adding it, we may want to think about decompression and how to implement it, see #35

amontoison:
@gdalle 
I will need your help to find the correct size to initialize `first_visit_to_tree`.
Can we do better than `first_visit_to_tree = fill((0, 0), nnz(g))`?
I don't know if we can have cases with **#trees == #edges**.

I have a functional function `acyclic_coloring` but it doesn't return a correct coloring yet.
```julia
using SparseMatrixColorings

A = [
   1  2  0  0  0  0  3  0  0  0
   2  4  5  0  6  0  0  0  0  0
   0  5  7  8  0  9  0  0  0  0
   0  0  8  10 0  0  0  0  0  11
   0  6  0  0  12 13 0  14 0  0
   0  0  9  0  13 15 0  0  16 0
   3  0  0  0  0  0  17 18 0  0
   0  0  0  0  14 0  18 19 20 0
   0  0  0  0  0  16 0  20 21 22
   0  0  0  11 0  0  0  0  22 23
]

g = SparseMatrixColorings.Graph(A)

colors = SparseMatrixColorings.acyclic_coloring(g, SparseMatrixColorings.NaturalOrder())

@test colors == [1, 2, 1, 2, 1, 3, 2, 3, 2, 1]
```

amontoison:
I checked the code in JuMP, and they use the number of edges:
https://github.com/jump-dev/MathOptInterface.jl/blob/master/src/Nonlinear/ReverseAD/Coloring/Coloring.jl#L184-L187.
They also initialize disjoint_sets with the number of edges instead of dynamically using push! like I do.

gdalle:
I merged the new `ColoringProblem` interface from #45 into this PR

gdalle:
I merged a few more PRs into this branch, including one that cleans up the examples from the paper figures. Interestingly, Figures 6.1 from "What" and Figure 4 from "Efficient" have the same matrix but different colorings, so it's expected that you wouldn't find both

amontoison:
> I merged a few more PRs into this branch, including one that cleans up the examples from the paper figures. Interestingly, Figures 6.1 from "What" and Figure 4 from "Efficient" have the same matrix but different colorings, so it's expected that you wouldn't find both

I remarked that but the coloring of Figure 6.1 just an example of an acyclic coloring, it's not related to any code so I don't obtain it.
However it's a nice example for testing decompression.

PR #38, opened 2024-08-06
Introduce coloring results

gdalle:
> [!WARNING]
> Breaking change: bumping version to v0.4.0

**Source**

- [x] Introduce a new "detailed" interface with three functions: `column_coloring_detailed`, `row_coloring_detailed` and `symmetric_coloring_detailed`. Make the ADTypes interface fall back on those (we'll see once this stabilizes if it is worth upstreaming).
- [x] Let the detailed coloring functions return an `AbstractColoringResult`, which must then be used in decompression routines. A minima, this object contains the `color` vector and the `group` vector. There is also one specific implementation for `SparseMatrixCSC`, where decompression indices are stored.
- [x] Anticipate generalizations of coloring results to 1) bidirectional coloring and 2) indirect recovery with linear systems, thanks to a flexible set of type parameters on `AbstractColoringResult{partition,symmetric,decompression}`. 
- [x] Merge `decompress_columns`, `decompress_row` and `decompress_symmetric` into a single symbol `decompress`, using dispatch on the result type to figure out the right algorithm.
- [x] Relax tests on same sparsity pattern, which added significant overhead.

**Docs**

- [x] Update docstrings and public symbols

**Tests**

- [x] Adjust to new framework
- [x] Test in-place decompression too

---

@amontoison this should be very easy to adapt to sparse GPU matrices


amontoison:
Should we add `ncolors` in `SimpleColoringResult` by default?

gdalle:
You can get it as `length(get_groups(coloring_result))`.

Don't mind the CI, I'm currently debugging something that only occurs on LTS

gdalle:
@amontoison any idea how this could ever happen for `A::SparseMatrixCSC`? It's the reason why CI fails, apparently only on 1.6
image

gdalle:
Okay apparently it's caused by https://github.com/JuliaSparse/SparseArrays.jl/issues/44 and fixed in https://github.com/JuliaLang/julia/pull/40523

gdalle:
@amontoison this is nearly done, care to take a look?

amontoison (approved):

amontoison:
LGTM!

Issue #68, opened 2024-08-12
Optimize decompression by substitution for sparse and dense


gdalle:
Once we have solved #44

PR #70, opened 2024-08-12
Update the structure TreeSet

amontoison:
Related to #68


gdalle:
This is superseded by the result type refactor in #71. 
I put the complexity in the result constructors rather than the `TreeSet`, that way we can still have a fast underlying coloring if we want to.

PR #73, opened 2024-08-13
More optimization for the decompression of the acyclic coloring

amontoison:
Related to #68 
- Efficient implementation of `TreeSetColoringResult`


amontoison:
@gdalle 
The PR is ready.
We have the optimal structure for decompression and, in my opinion, the best method to compute it.
The only potential improvement could be to apply tricks to minimize dynamic allocations and reuse existing storage.
We can further optimize the code if, one day, someone encounters an issue with a very large graph.

PR #83, opened 2024-08-14
Add option to decompress into a triangle only

gdalle:
- [x] Add `uplo = :U/:L/:UL` parameter to `decompress!` and `decompress_single_color!` to allow picking the triangle for symmetric matrices
- [x] Implement the corresponding methods
- [x] Add tests for full decompression and for single color, structure into subtests

### DIFF
--- a/src/decompression.jl
+++ b/src/decompression.jl
@@ -124,7 +124,7 @@ end
 """
     decompress!(
         A::AbstractMatrix, B::AbstractMatrix,
-        result::AbstractColoringResult,
+        result::AbstractColoringResult, [uplo=:UL]
     )
 
 Decompress `B` in-place into `A`, given a coloring `result` of the sparsity pattern of `A`.
@@ -132,6 +132,8 @@ The out-of-place alternative is [`decompress`](@ref).
 
 Compression means summing either the columns or the rows of `A` which share the same color.
 It is done by calling [`compress`](@ref).
+
+For `:symmetric` coloring results (and for those only), an optional positional argument `uplo in (:U, :L, :UL)` can be passed to specify which triangle of the matrix `A` should be updated: the upper one, the lower one, or both.
 
 !!! note
     In-place decompression is faster when `A isa SparseMatrixCSC`.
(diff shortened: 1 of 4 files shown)


gdalle:
@amontoison you asked why `uplo` is not a kwarg. The reason is that I want to be able to dispatch on whether it is present or not, and you cannot dispatch on `kwargs` (not even on their presence or absence).

My reasoning is that between the following methods, the first one is more efficient (leads to predictable memory accesses) and should thus be preferred when `uplo` is not specified. What do you think?

```julia

function decompress!(
    A::SparseMatrixCSC{R}, B::AbstractMatrix{R}, result::StarSetColoringResult
) where {R<:Real}
    @compat (; S, compressed_indices) = result
    check_same_pattern(A, S)
    nzA = nonzeros(A)
    for k in eachindex(nzA, compressed_indices)
        nzA[k] = B[compressed_indices[k]]
    end
    return A
end

function decompress!(
    A::SparseMatrixCSC{R}, B::AbstractMatrix{R}, result::StarSetColoringResult, uplo::Symbol
) where {R<:Real}
    @compat (; S, compressed_indices) = result
    check_same_pattern(A, S)
    rvA = rowvals(A)
    nzA = nonzeros(A)
    for j in axes(S, 2)
        for k in nzrange(S, j)
            i = rvA[k]
            if in_triangle(i, j, uplo)
                nzA[k] = B[compressed_indices[k]]
            end
        end
    end
    return A
end
```

amontoison:
You will not gain a lot if you do that:
```julia
function decompress!(
    A::SparseMatrixCSC{R}, B::AbstractMatrix{R}, result::StarSetColoringResult; uplo::Symbol=:F
) where {R<:Real}
    @compat (; S, compressed_indices) = result
    check_same_pattern(A, S)
    nzA = nonzeros(A)
    if uplo == :F
      for k in eachindex(nzA, compressed_indices)
          nzA[k] = B[compressed_indices[k]]
      end
    else
      rvA = rowvals(A)
      for j in axes(S, 2)
        for k in nzrange(S, j)
          i = rvA[k]
          if in_triangle(i, j, uplo)
            nzA[k] = B[compressed_indices[k]]
          end
        end
      end
    end
    return A
end
``

amontoison:
The issue is that sometimes `uplo` is a kwarg and sometimes not.

gdalle:
> You will not gain a lot if you do that:

True but with dispatch I avoid compiling the useless part of the code.

> The issue is that sometimes uplo is a kwarg and sometimes not.

There was one spot where it was (erroneously) a kwarg, I fixed it in #85. I think now it should be a positional argument everywhere.

amontoison:
You code too fast for me. It's perfect in that case.

PR #85, opened 2024-08-14
Replace `:UL` with `:F` to select a triangle

gdalle:
- Replace the default value of `uplo` with `:F`, which is apparently common in numerical linear algebra.

### DIFF
--- a/src/decompression.jl
+++ b/src/decompression.jl
@@ -513,7 +513,10 @@ end
 ## MatrixInverseColoringResult
 
 function decompress!(
-    A::AbstractMatrix{R}, B::AbstractMatrix{R}, result::LinearSystemColoringResult; uplo=:F
+    A::AbstractMatrix{R},
+    B::AbstractMatrix{R},
+    result::LinearSystemColoringResult,
+    uplo::Symbol=:F,
 ) where {R<:Real}
     @compat (;
         S, color, strict_upper_nonzero_inds, T_factorization, strict_upper_nonzeros_A
--- a/src/decompression.jl
+++ b/src/decompression.jl
@@ -124,7 +124,7 @@ end
 """
     decompress!(
         A::AbstractMatrix, B::AbstractMatrix,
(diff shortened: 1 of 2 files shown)


src/decompression.jl
```diff
@@ -133,7 +133,7 @@ The out-of-place alternative is [`decompress`](@ref).
 Compression means summing either the columns or the rows of `A` which share the same color.
 It is done by calling [`compress`](@ref).
 
-For `:symmetric` coloring results (and for those only), an optional positional argument `uplo in (:U, :L, :UL)` can be passed to specify which triangle of the matrix `A` should be updated: the upper one, the lower one, or both.
+For `:symmetric` coloring results (and for those only), an optional positional argument `uplo in (:U, :L, :F)` can be passed to specify which triangle of the matrix `A` should be updated: the Upper one, the Lower one, or the Full matrix.
```
amontoison, line 136: Keep things in lowercase?

src/decompression.jl
```diff
@@ -197,7 +197,7 @@ Decompress the vector `b` corresponding to color `c` in-place into `A`, given a
 - If `result` comes from a `:nonsymmetric` structure with `:row` partition, this will update the rows of `A` that share color `c` (whose sum makes up `b`).
 - If `result` comes from a `:symmetric` structure with `:column` partition, this will update the coefficients of `A` whose value is deduced from color `c`.
 
-For `:symmetric` coloring results (and for those only), an optional positional argument `uplo in (:U, :L, :UL)` can be passed to specify which triangle of the matrix `A` should be updated: the upper one, the lower one, or both.
+For `:symmetric` coloring results (and for those only), an optional positional argument `uplo in (:U, :L, :F)` can be passed to specify which triangle of the matrix `A` should be updated: the upper one, the lower one, or both.
```
amontoison, line 200: Update the docstring to be similar to the previous one that you updated.

gdalle: done, thanks

PR #90, opened 2024-08-15
Fix star decompression into `SparseMatrixCSC` triangle

gdalle:
- The previous version assumed that the sparsity patterns of `S` and `A` were aligned. If `A` is just one of the triangles, we need to count its indices separately to map them onto `compressed_indices`.
- Adjust tests to decompress into `similar(triu(A))` instead of `similar(A)`

### DIFF
--- a/test/allocations.jl
+++ b/test/allocations.jl
@@ -54,7 +54,7 @@ function test_noallocs_decompression(
     end
     @testset "Triangle decompression" begin
         if structure == :symmetric
-            bench1_triangle = @be similar(A) decompress!(_, B, result, :U) evals = 1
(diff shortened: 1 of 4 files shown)


test/allocations.jl
```diff
@@ -54,7 +54,7 @@ function test_noallocs_decompression(
     end
     @testset "Triangle decompression" begin
         if structure == :symmetric
-            bench1_triangle = @be similar(A) decompress!(_, B, result, :U) evals = 1
+            bench1_triangle = @be similar(triu(A)) decompress!(_, B, result, :U) evals = 1
```
amontoison, line 57: Tu peux tester avec `tril(A)` aussi?

amontoison (approved):

PR #96, opened 2024-09-06
Update interface.jl

amontoison:
@gdalle


gdalle:
I think this is a misunderstanding: here `decompression` denotes a variable which contains a symbol (equal to `:direct` or `:substitution`), not the symbol `:decompression`.

amontoison:
You're right I got mixed up with something else. I tried `GreedyColoringAlgorithm{substitution}(NaturalOrder())
` and opened the PR.

PR #107, opened 2024-09-22
Manual graph transposition

amontoison:


gdalle:
I am not a fan of reimplementing what's already in a standard library just to shave off one allocation (the new nzval, if I understand correctly). Are you sure there is no other way?

amontoison:
I don't see another solution.
But It's just a routine of 50 lines of code. We don't reimplement a sparse Cholesky so I don't see the problem.

amontoison:
@gdalle What is the command to apply the formater on the code?

gdalle:
JuliaFormatter.format with a path, or just control shift I in VSCode.
Okay for this custom transposition PR but can you add unit tests?

amontoison:
@gdalle Can you approve it such that I don't rebase it after #112?

gdalle:
Actually I wanted to work some more on this one, since the other PR is very short I'll take care of rebasing

gdalle:
I don't really like this handwritten loop with obscure logic, even though you documented it well I still struggle to understand it. I opened #114 with a much more elegant 3-line implementation, which should still shave off all unneeded allocations but one (haven't benchmarked though). I suggest we merge that one and possibly revisit this afterwards.

gdalle:
Also, some of the benchmarks show degradation but I think that's just because they're incredibly noisy. Will try to fix them

amontoison:
> I don't really like this handwritten loop with obscure logic, even though you documented it well I still struggle to understand it. I opened #114 with a much more elegant 3-line implementation, which should still shave off all unneeded allocations but one (haven't benchmarked though). I suggest we merge that one and possibly revisit this afterwards.

If you propose, don't merge it!
I don't find your 3-line implementation elegant.
I prefer this PR and would like this to be merged.

gdalle:
> If you propose, don't merge it!

You're right, I'm sorry. I merged mine first because the benchmarks showed virtually no difference, but I should have waited for you to weigh in.

> I don't find your 3-line implementation elegant.

We obviously have diverging opinions on this, so let's judge from the benchmarks? I have merged `main` into this PR, so they will show improvements of your method against mine.

> It's a very classic CSR --> CSC implementation, if you don't understand it, it doesn't mean that you should do your own version.

Precisely because it is so classic, my assumption is that there is a version of it in SparseArrays that is better than ours, and we should just reuse it instead of adding complexity to our code base.

gdalle (approved):

PR #113, opened 2024-09-26
Remove dynamic allocations in group_by_color

amontoison:
@gdalle It seems that we have unexpected allocations in `ColumnResult` and `StarSetResult`.
I will investigate.


gdalle:
You may wanna use Chairmarks to make sure that JIT-compilation and other effects like that don't affect your measurement? I did the same just above

gdalle:
I think what you're seeing is `group_by_color`

amontoison:
Oh, you're right. I thought the structure had already been computed, and that group_by_color was just a getter.

gdalle (approved):

PR #114, opened 2024-09-26
Improve graph structure a little bit

gdalle:
- Store size information in the `Graph` structure
- Define `nb_vertices` and `nb_edges` in lieu of `length` and `nnz` to avoid confusion
- Redefine `transpose_graph(::AbstractMatrix)` from #107 as `Base.transpose(::Graph)`
- Use a simpler implementation which does just one allocation too many (a new vector of `nzval`), instead of a custom handwritten loop (at least I think so)
- Add better transposition tests

### DIFF
--- a/src/graph.jl
+++ b/src/graph.jl
@@ -3,8 +3,8 @@
 """
     Graph{loops,T}
 
-Undirected graph structure stored in Compressed Sparse Column (CSC) format.
-Note that the underyling CSC may not be square, so the term "graph" is slightly abusive.
+Store a sparse matrix (in CSC) without its values, keeping only the pattern of nonzeros.
+It can be seen as a graph mapping columns to rows, hence the name `Graph`.
 
 The type parameter `loops` must be set to:
 - `true` if coefficients `(i, i)` present in the CSC are counted as edges in the graph (e.g. for each half of a bipartite graph)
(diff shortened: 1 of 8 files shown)


src/graph.jl
```diff
@@ -69,6 +79,17 @@ end
...
 minimum_degree(g::Graph) = minimum(Base.Fix1(degree, g), vertices(g))
 
+"""
+    transpose(g::Graph)
+
+Return a [`Graph`](@ref) corresponding to the transpose of (the underlying matrix of) `g`.
+"""
+function Base.transpose(g::Graph{loops,T}) where {loops,T}
+    S = SparseMatrixCSC{T,T}(g.m, g.n, g.colptr, g.rowval, g.rowval)
+    Sᵀ = convert(SparseMatrixCSC, transpose(S))  # TODO: use ftranspose! without segfault?
```
gdalle, line 89: @amontoison my initial idea was to use [`ftranspose!`](https://docs.julialang.org/en/v1.11.0-rc4/stdlib/SparseArrays/#SparseArrays.ftranspose!) as follows:
```julia
new_colptr = Vector{T}(undef, g.m + 1)
new_rowval = similar(g.rowval) 
Sᵀ = SparseMatrixCSC{T,T}(g.n, g.m, new_colptr, new_rowval, g.rowval)
ftranspose!(Sᵀ, S, identity)
```
This would avoid any non-necessary allocation, but unfortunately it causes segfaults. Presumably because `Sᵀ` and `S` share their vector of `nzval`, which I thought wouldn't be a problem but it is.

amontoison:
@gdalle Please wait that I have a have a look before merging...

PR #123, opened 2024-09-27
Clearly separate graphs from the sparsity pattern

gdalle:
**Idea**

Create two clearly distinct graph structures (`AdjacencyGraph` and `BipartiteGraph`) and distinguish them from the container for the sparsity pattern.

**Benefits**

Now:
- Get rid of the weird `Graph{true}` / `Graph{false}` which wasn't even a graph in the rectangular case.
- Allow storing more stuff in the graph structure if we want (typically the number of edges as you suggested).

In the future:
- Gain flexibility to implement graphs based on different pattern types (for other structured matrices like banded), just by making the pattern fields parametric.

**Changes**

- Rename `Graph` into `SparsePatternCSC`, which is exactly a `SparseMatrixCSC` without `nzval`. Remove definitions of `neighbors` and friends, keep only linear algebra stuff there.
- Create a proper `AdjacencyGraph` based on one `SparsePatternCSC`, put the `neighbors` logic there. It is the pendant to `BipartiteGraph`.
- Count edges properly in the `AdjacencyGraph` (divide by two to account for symmetry).
- Remove the constructors `adjacency_graph` and `bipartite_graph`.

### DIFF
--- a/src/graph.jl
+++ b/src/graph.jl
@@ -218,7 +218,9 @@ function neighbors(bg::BipartiteGraph, ::Val{side}, v::Integer) where {side}
     return view(rowvals(S), nzrange(S, v))
 end
 
-degree(bg::BipartiteGraph, ::Val{side}, v::Integer) where {side} = length(neighbors(bg, v))
+function degree(bg::BipartiteGraph, ::Val{side}, v::Integer) where {side}
+    return length(neighbors(bg, Val(side), v))
+end
 
 function maximum_degree(bg::BipartiteGraph, ::Val{side}) where {side}
     return maximum(v -> degree(bg, Val(side), v), vertices(bg, Val(side)))
--- a/src/graph.jl
+++ b/src/graph.jl
@@ -1,14 +1,9 @@
 ## Standard graph
 
 """
-    Graph{loops,T}
+    SparsePatternCSC{Ti<:Integer}
 
 Store a sparse matrix (in CSC) without its values, keeping only the pattern of nonzeros.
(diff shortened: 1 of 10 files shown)


src/graph.jl
```diff
@@ -129,32 +78,128 @@ function Base.transpose(g::Graph{loops,T}) where {loops,T}
...
+    return d
+end
+
+function nb_edges(g::AdjacencyGraph)
+    S = pattern(g)
+    ne = 0
+    for j in vertices(g)
+        for k in nzrange(S, j)
+            i = rowvals(S)[k]
+            if i != j
```
amontoison, line 140: Maybe just check that `i < j`. You don't need `ne ÷ 2` at the end with this modification.

amontoison (changes requested):
@gdalle
Excellent PR! I really like this new structure.
I was not a big fan of the `Graph` with the attribute `loops`.
Aside from a small comment about the number of edges for `AdjacencyGraph`, the code looks good to me.
Could I request the name `SparsityPatternCSC` instead of `SparsePatternCSC`?
It’s even more explicit from my point of view.

Excellent job!!!

amontoison (approved):

amontoison:
It should be easily to remove the allocation of S'  [here](https://github.com/gdalle/SparseMatrixColorings.jl/blob/main/src/result.jl#L153) now.

gdalle:
Indeed that was the next step of the plan, I have opened https://github.com/gdalle/SparseMatrixColorings.jl/issues/126 to keep track

Issue #15, opened 2024-05-27
More options for orders


gdalle:
To compare with ColPack (https://dl.acm.org/doi/10.1145/2513109.2513110) we need the following orders:

- [x] Natural order (N)
- [x] Random order (R)
- [x] Largest first (LF)
- [x] Smallest last  (SL)
- [x] Incidence degree (ID)
- [x] Dynamic largest first (DLF)
- [ ] Saturation degree (SD) (not in ColPack)

gdalle:
Mostly solved in #125

PR #125, opened 2024-09-27
Degree-based orderings, take 2

gdalle:
This PR implements three additional orders from the ColPack paper: `SmallestLast`, `IncidenceDegree`, `DynamicLargestFirst` (new version of #18).

- [x] Implement degree-based orders for `AdjacencyGraph` and `BipartiteGraph`.
- [x] Test that they return correct results.

The big question is where to find small examples to check that we return the correct orders. This is especially tricky because ties are broken in an arbitrary manner.


amontoison:
You also implemented `DynamicDegreeBasedOrder`, right?

gdalle:
> You also implemented DynamicDegreeBasedOrder, right?

Yeah, basically these three new orders are variants of the same underlying paradigm, which is `DynamicDegreeBasedOrder`. So we don't need three different implems, just a single one with a few toggles here and there

amontoison:
gdalle, what is `SaturationDegree` in #15?
Is it the missing combination forward + up?

gdalle:
No, it's a very different one. I just re-read the section of the ColPack paper and I think we can skip it?

> Saturation Degree ordering, due to Brélaz [1979], is another known ordering technique for greedy coloring. It can be best viewed in contrast to the ID ordering technique: in a coloring algorithm that uses an ID ordering, the ith vertex is a vertex with the largest number of already colored adjacent vertices, whereas in an algorithm that uses an SD ordering it is a vertex that has the largest number of distinctly colored adjacent vertices. Clearly, the characterization of an SD ordering cannot be separated from the rest of the coloring algorithm, and the ordering is nonamenable for a linear-time implementation in the size of the graph. For these reasons we chose not to include it in ColPack. Turner [1988] has shown that using careful choice of data structures, a coloring algorithm that in effect employs an SD ordering can be implemented such that its complexity is `O(|E|log|V|)`. We have implemented such an algorithm and we will report results on it in the experiments discussed in Section 8.

gdalle:
@amontoison friendly bump, do you want us to try and merge this one?

gdalle:
Following our chat with @amontoison, I'm merging this with clear warnings that the new orders are experimental and need further tests. That way we can start experimenting.

PR #129, opened 2024-09-28
Merge two decompress! for star coloring

amontoison:
We also want the optimized version of `decompress!` for star coloring when the user specifies `uplo=:F` and not only when `uplo` is not specified.

### DIFF
--- a/src/decompression.jl
+++ b/src/decompression.jl
@@ -257,7 +257,7 @@ function in_triangle(i::Integer, j::Integer, uplo::Symbol)
         return true
     elseif uplo == :U
         return i <= j
-    else
+    else  # uplo == :L
         return i >= j
     end
 end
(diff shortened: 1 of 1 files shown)


src/decompression.jl
```diff
@@ -256,9 +256,9 @@ function in_triangle(i::Integer, j::Integer, uplo::Symbol)
     if uplo == :F
         return true
     elseif uplo == :U
-        return i <= j
-    else
-        return i >= j
+        return i ≤ j
```
gdalle, line 259:
```suggestion
        return i <= j
```
we didn't use unicode anywher else, so I'd prefer us to remain consistent

src/decompression.jl
```diff
@@ -256,9 +256,9 @@ function in_triangle(i::Integer, j::Integer, uplo::Symbol)
...
+    else  # uplo == :L
+        return i ≥ j
```
gdalle, line 261:
```suggestion
        return i >= j
```

src/decompression.jl
```diff
@@ -434,35 +434,32 @@ function decompress_single_color!(
...
+                i = rvS[k]
+                if in_triangle(i, j, uplo)
+                    l += 1
+                    nzA[l] = B[compressed_indices[k]]
+                end
             end
         end
+        @assert l == nnz(A)
     end
-    @assert l == length(nonzeros(A))
```
gdalle, line 465: This was here for a very specific reason. I don't rember the exact discussion around it but the gist is that we don't have the guarantee `nnz(A) == nonzeros(A)`, so checking against `length(nonzeros(A))` was safer.

gdalle (changes requested):

src/decompression.jl
gdalle: We should probably add a comment explaining that though

amontoison: I remember now. Should we remove this `@assert`?

gdalle: I would keep it, it's a guardrail against simple user mistakes which would otherwise silently return the wrong result. Typically if the user passes the whole matrix in SparseMatrixCSC but requires one of the two triangles, the filling will be all wrong

amontoison:
@gdalle It should be good now :)

gdalle (approved):

PR #130, opened 2024-09-28
Decompression of acyclic coloring with SparseMatrixCSC

### LINKED ISSUE #93: Decompression of acyclic coloring with SparseMatrixCSC
The current code is quite optimal, but we could optimize further for `SparseMatrixCSC` by storing [here](https://github.com/gdalle/SparseMatrixColorings.jl/blob/main/src/decompression.jl#L502) four values `(i, j, k1, k2)`, where `k1` is the position in `A.nzval` of `A[i,j]` and `k2` is the position of `A[j,i]` in `A.nzval`.

We should do the same thing for diagonal coefficients [here](https://github.com/gdalle/SparseMatrixColorings.jl/blob/main/src/decompression.jl#L492).

The current issue is that I don't know how to use this information if we just want to update one triangle of the Hessian.
The positions in `A.nzval` are not valid anymore.

amontoison:
close #93

- Add the following method:
```julia
decompress!(A::SparseMatrixCSC{R}, B::AbstractMatrix{R}, result::TreeSetColoringResult, uplo::Symbol=:F)
   ...
end
```

### DIFF
--- a/src/decompression.jl
+++ b/src/decompression.jl
@@ -460,8 +460,6 @@ end
 
 ## TreeSetColoringResult
 
-# TODO: add method for A::SparseMatrixCSC
-
 function decompress!(
     A::AbstractMatrix, B::AbstractMatrix, result::TreeSetColoringResult, uplo::Symbol=:F
 )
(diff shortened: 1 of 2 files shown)


amontoison:
@gdalle I spend a few hours to implement an optimized version and it's 2-4x time slower :skull:

gdalle:
I think you're reading the benchmarks wrong? The last column is `main` over your branch, so your method is actually faster?

amontoison:
Yep, you are right. I don't know how to read the benchmarks 🤪

amontoison:
@gdalle Can you review this PR if you find time?

gdalle:
Sure, I'll do it tomorrow

src/decompression.jl
```diff
@@ -504,6 +502,106 @@ function decompress!(
...
+        for (i, j) in reverse_bfs_orders[k]
+            counter += 1
+            val = B[i, color[j]] - buffer_right_type[i]
+            buffer_right_type[j] = buffer_right_type[j] + val
+
+            #! format: off
+            # A[i,j] is in the lower triangular part of A
+            if in_triangle(i, j, :L)
+                # uplo = :L or uplo = :F
+                # A[i,j] is stored at index_ij = (A.colptr[j+1] - offset_L) in A.nzval
```
gdalle, line 570: Why did you choose to subtract the lower triangle offsets but add the upper triangle ones?

src/decompression.jl
```diff
@@ -504,6 +502,106 @@ function decompress!(
...
+        end
+
+        for (i, j) in reverse_bfs_orders[k]
+            counter += 1
+            val = B[i, color[j]] - buffer_right_type[i]
+            buffer_right_type[j] = buffer_right_type[j] + val
+
+            #! format: off
+            # A[i,j] is in the lower triangular part of A
+            if in_triangle(i, j, :L)
```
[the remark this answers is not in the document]
gdalle: I'm asking because these nested conditions with `if uplo != ...` make it a bit harder to reason about the code, at least for me

src/decompression.jl
```diff
@@ -504,6 +502,106 @@ function decompress!(
...
+            nzA[nzind] = B[i, color[i]]
+        end
+    else  # uplo == :F
+        for (k, i) in enumerate(diagonal_indices)
+            nzind = diagonal_nzind[k]
+            nzA[nzind] = B[i, color[i]]
+        end
+    end
+
+    # Index of offsets in lower_triangle_offsets and upper_triangle_offsets
```
gdalle, line 552: The counter evolves in the same order as the reverse BFS?

src/result.jl
```diff
@@ -373,6 +403,36 @@ function TreeSetColoringResult(
...
+                    #! format: off
+                    # S[i,j] is in the lower triangular part of S
+                    if in_triangle(i, j, :L)
+                        # uplo = :L or uplo = :F
+                        # S[i,j] is stored at index_ij = (S.colptr[j+1] - offset_L) in S.nzval
+                        lower_triangle_offsets[index_offsets] = length(col_j) - searchsortedfirst(col_j, i) + 1
+
+                        # uplo = :U or uplo = :F
+                        # S[j,i] is stored at index_ji = (S.colptr[i] + offset_U) in S.nzval
+                        upper_triangle_offsets[index_offsets] = searchsortedfirst(col_i, j)::Int - 1
```
gdalle, line 423:
```suggestion
                        upper_triangle_offsets[index_offsets] = searchsortedfirst(col_i, j) - 1
```

src/result.jl
```diff
@@ -373,6 +403,36 @@ function TreeSetColoringResult(
...
+
+                    # S[i,j] is in the upper triangular part of S
+                    else
+                        # uplo = :U or uplo = :F
+                        # S[i,j] is stored at index_ij = (S.colptr[j] + offset_U) in S.nzval
+                        upper_triangle_offsets[index_offsets] = searchsortedfirst(col_j, i)::Int - 1
```
gdalle, line 429:
```suggestion
                        upper_triangle_offsets[index_offsets] = searchsortedfirst(col_j, i) - 1
```

src/decompression.jl
amontoison: Because it's the only way to have the same offsets with a matrix that contains both triangles.

src/decompression.jl
amontoison: Yes, but you perform a reverse BFS for each tree you have. So it's a cumulative counter.

src/result.jl
amontoison: `JET` is not happy without that. I don't understand why.
It seems related to a type piracy of `DataStructures.jl`.

src/result.jl
amontoison: Same comment as above.

src/decompression.jl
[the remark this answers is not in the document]
amontoison: Because it depends on which triangle `A[i,j]` is in (2 combinations) and which triangle the user wants (2 combinations).
These two values are independent, so we need four branches.
I agree that it's not straightforward, but I couldn't find a better solution.

src/result.jl
gdalle: Apparently the type of `col_i` is not properly inferred, which is weird cause it works for `col_j`. Let's leave it like that

gdalle (approved):

PR #132, opened 2024-10-02
Coloring and decompression for structured matrices

### LINKED ISSUE #64: Optimized decompression for specific matrix types
At the moment, our only optimized decompression is for `SparseMatrixCSC` in `:direct` mode: we store a vector of `compressed_indices` such that `nonzeros(A) = vec(B)[compressed_indices]`.
We can probably find a similar optimization for `:substitution` mode.

What do we want to do for other matrix types, like:

- the ones from `LinearAlgebra`: `Bidiagonal`, `Tridiagonal`, `Symmetric`, etc.
- [BandedMatrices.jl](https://github.com/JuliaLinearAlgebra/BandedMatrices.jl), [BlockBandedMatrices.jl](https://github.com/JuliaLinearAlgebra/BlockBandedMatrices.jl)

It would be rather tiring to find optimal decompression methods for each of these. My proposal (as a first step) would be to always have a `SparseMatrixCSC` buffer into which we decompress, and then copy the `A_buffer::SparseMatrixCSC` into `A::SomeWeirdMatrix`.
Essentially, it's easier to implement fast copy from `SparseMatrixCSC` than fast decompression.

Related:

- #65 
- #44

### LINKED ISSUE #65: Optimized coloring for specific matrix types
There are some matrix types for which we can directly compute the optimal coloring, like (block-)banded matrices.

[ArrayInterface.jl](https://github.com/JuliaArrays/ArrayInterface.jl) does this already but only for column coloring. We could reuse some of that code (if the license is right) and generalize it to other colorings (row, symmetric, bidirectional).

Related:

- #64

gdalle:
Fixes #64 and fixes #65 by implementing structured coloring & decompression for various structured matrix types:

| Package | Type | Coloring | Decomp. | Test perf |
|---|---|---|---|---|
| LinearAlgebra | `Diagonal` | ✅ | ✅ | ✅ |
| LinearAlgebra | `Bidiagonal` | ✅ | ✅ | ✅ |
| LinearAlgebra | `Tridiagonal` | ✅ | ✅ | ✅ |
| LinearAlgebra | `SymTridiagonal` |  |  |  |
| BandedMatrices | `BandedMatrix` | ✅ | ✅ |  |
| BlockBandedMatrices | `BlockBandedMatrix` | ✅ |  |  |
| BlockBandedMatrices | `BandedBlockBandedMatrix` | ✅ |  |  |

**General philosophy**

Keep the same result type (`ColumnColoringResult` / `RowColoringResult`).

- Upside: if the user tries to decompress into something else, like a `SparseMatrixCSC`, it will still be very fast
- Downside: a few needless allocations after coloring, but unlike decompression they are a one-time cost.

**Remaning issues**

For `SymTridiagonal` I'm not sure what to do since it is not mutable.

For decompression of BlockBandedMatrices, I still need to figure out https://github.com/JuliaArrays/BlockArrays.jl/issues/418. But it already works because the fallback algorithm relies on the nonzeros of the sparse encoding, so it does not try to access illegal indices outside of the bands.


gdalle:
Superseded by #139 (I screwed up with a `git rebase`)

PR #133, opened 2024-10-02
Store graph in result to allow generic matrices

### LINKED ISSUE #126: Store graph object in coloring result
This can help remove some allocations, e.g. for `S'` in row decompression

### LINKED ISSUE #131: Avoid forcing eltype equality between `A` and `B` in `decompress!`
This might mean we hit slow dispatches or error when promotion would have done the trick

gdalle:
Fix #126, fix #131

Before this PR, the `result` type would always store a `SparseMatrixCSC` and forget the initial matrix `A`. Now it stores both the initial matrix `A` and the `AdjacencyGraph` or `BipartiteGraph` used for coloring, which contains the sparse conversion. This has two main advantages:

- To prepare for row decompression we do not need to recompute a transpose, it is already done in the `BipartiteGraph`
- When decompressing out-of-place, we can instantiate `similar(A)` instead of `similar(S)`, which makes more sense and paves the way for structured matrices

I also remove some eltype constraints on `A` and `B` which were not necessary. For instance, if `eltype(B) == Float32` and `eltype(A) == Float64`, we can decompress from `B` to `A` without issue.

### DIFF
--- a/src/decompression.jl
+++ b/src/decompression.jl
@@ -377,7 +377,7 @@ function decompress!(
     @compat (; star, hub, spokes) = star_set
     S = result.ag.S
     uplo == :F && check_same_pattern(A, S)
-    A .= zero(eltype(A))
+    fill!(A, zero(eltype(A)))
     for i in axes(A, 1)
         if !iszero(S[i, i])
             A[i, i] = B[i, color[i]]
@@ -469,7 +469,7 @@ function decompress!(
     S = result.ag.S
     uplo == :F && check_same_pattern(A, S)
     R = eltype(A)
-    A .= zero(R)
+    fill!(A, zero(R))
 
     if eltype(buffer) == R
         buffer_right_type = buffer
@@ -519,7 +519,7 @@ function decompress!(
 
     # TODO: for some reason I cannot use ldiv! with a sparse QR
     strict_upper_nonzeros_A = T_factorization \ vec(B)
-    A .= zero(eltype(A))
+    fill!(A, zero(eltype(A)))
     for i in axes(A, 1)
         if !iszero(S[i, i])
             A[i, i] = B[i, color[i]]
--- a/src/decompression.jl
+++ b/src/decompression.jl
@@ -115,9 +115,8 @@ true
 - [`ColoringProblem`](@ref)
 - [`AbstractColoringResult`](@ref)
 """
-function decompress(B::AbstractMatrix{R}, result::AbstractColoringResult) where {R<:Real}
-    @compat (; S) = result
-    A = respectful_similar(S, R)
+function decompress(B::AbstractMatrix, result::AbstractColoringResult)
+    A = respectful_similar(result.A, eltype(B))
     return decompress!(A, B, result)
 end
 
@@ -264,12 +263,11 @@ end
 
 ## ColumnColoringResult
 
-function decompress!(
-    A::AbstractMatrix{R}, B::AbstractMatrix{R}, result::ColumnColoringResult
-) where {R<:Real}
-    @compat (; S, color) = result
+function decompress!(A::AbstractMatrix, B::AbstractMatrix, result::ColumnColoringResult)
+    @compat (; color) = result
+    S = result.bg.S2
     check_same_pattern(A, S)
-    A .= zero(R)
+    fill!(A, zero(eltype(A)))
     rvS = rowvals(S)
     for j in axes(S, 2)
         cj = color[j]
@@ -282,9 +280,10 @@ function decompress!(
 end
 
 function decompress_single_color!(
-    A::AbstractMatrix{R}, b::AbstractVector{R}, c::Integer, result::ColumnColoringResult
-) where {R<:Real}
-    @compat (; S, group) = result
+    A::AbstractMatrix, b::AbstractVector, c::Integer, result::ColumnColoringResult
+)
+    @compat (; group) = result
+    S = result.bg.S2
     check_same_pattern(A, S)
     rvS = rowvals(S)
     for j in group[c]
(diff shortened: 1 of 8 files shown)


src/decompression.jl
```diff
@@ -115,9 +115,8 @@ true
 - [`ColoringProblem`](@ref)
 - [`AbstractColoringResult`](@ref)
 """
-function decompress(B::AbstractMatrix{R}, result::AbstractColoringResult) where {R<:Real}
-    @compat (; S) = result
-    A = respectful_similar(S, R)
+function decompress(B::AbstractMatrix, result::AbstractColoringResult)
+    A = respectful_similar(result.A, eltype(B))
```
gdalle, line 119: more coherent `similar`

src/decompression.jl
```diff
@@ -264,12 +263,11 @@ end
...
-function decompress!(
-    A::AbstractMatrix{R}, B::AbstractMatrix{R}, result::ColumnColoringResult
-) where {R<:Real}
-    @compat (; S, color) = result
+function decompress!(A::AbstractMatrix, B::AbstractMatrix, result::ColumnColoringResult)
+    @compat (; color) = result
+    S = result.bg.S2
     check_same_pattern(A, S)
-    A .= zero(R)
+    fill!(A, zero(eltype(A)))
```
gdalle, line 270: I recently discovered that `fill!(A, val)` is much more efficient than `A .= val`

src/graph.jl
```diff
@@ -81,6 +94,15 @@ function Base.transpose(S::SparsityPatternCSC{T}) where {T}
     return SparsityPatternCSC{T}(n, m, B_colptr, B_rowval)
 end
 
+# copied from SparseArrays.jl
+function Base.getindex(S::SparsityPatternCSC, i0::Integer, i1::Integer)
```
gdalle, line 98: This is useful in symmetric colorings when we check the diagonal coefficients with `iszero(S[i, i])`

src/result.jl
```diff
@@ -141,16 +145,17 @@ $TYPEDFIELDS
...
+       AbstractColoringResult{:nonsymmetric,:row,:direct}
+    A::M
+    bg::G
     color::Vector{Int}
     group::Vector{Vector{Int}}
     compressed_indices::Vector{Int}
 end
 
-function RowColoringResult(S::SparseMatrixCSC, color::Vector{Int})
-    Sᵀ = convert(SparseMatrixCSC, transpose(S))
```
gdalle, line 153: This is the transposition we avoid by reusing the graph

src/graph.jl
gdalle: Original code can be inspected with
```julia
julia> using SparseArrays; @edit spzeros(2, 2)[1, 2]
```

src/decompression.jl
amontoison: On devrait rzjouter un petit check avec le promote_type pour être certains que ça marche.

src/graph.jl
```diff
@@ -24,6 +24,19 @@ end
...
 
 Base.size(S::SparsityPatternCSC) = (S.m, S.n)
+
+function Base.size(S::SparsityPatternCSC, d::Integer)
+    if d == 1
+        return S.m
+    elseif d == 2
+        return S.n
+    else
+        return 1
```
amontoison, line 34:
```suggestion
        error("The dimension must be 1 or 2.")
```

amontoison:
Except two small comments, it's LGTM.

src/graph.jl
gdalle: Je pensais aussi, mais en fait les arrays natifs Julia se comportent comme ça :

```julia
julia> A = rand(2, 3);

julia> size(A, 1)
2

julia> size(A, 2)
3

julia> size(A, 3)
1

julia> size(A, 1000)
1
```

src/decompression.jl
gdalle: En vrai si ça marche pas Julia nous lancera une exception sur la promotion et ce sera tout aussi clair qu'un message d'erreur manuel non ?

src/graph.jl
amontoison: C'est aussi le cas avec une SparseMatrixCSC?
Notre type c'est une SparseArrays.AbstractSparseMatrixCSC dans l'idée

src/decompression.jl
amontoison: Pas faux. On peut laisser comme ça

src/graph.jl
gdalle: J'ai repris la définition de `Base`, elle est encore plus simple:
```
Base.size(S::SparsityPatternCSC, d) = d::Integer <= 2 ? size(S)[d] : 1
```

gdalle:
I think both those comments are resolved?

src/graph.jl
```diff
@@ -24,17 +24,7 @@ end
...
-    if d == 1
-        return S.m
-    elseif d == 2
-        return S.n
-    else
-        return 1
-    end
-end
-
+Base.size(S::SparsityPatternCSC, d) = d::Integer <= 2 ? size(S)[d] : 1
```
amontoison, line 27: 1 <= d::Integer <= 2 🙂

amontoison: À part si tu laisses Julia faire une erreur volontairement

src/graph.jl
amontoison: Ok j'avais loupé ton commentaire!

amontoison (approved):

src/graph.jl
gdalle: Funnily enough, this is exactly what they do in `Base`, I literally copy-pasted this line
https://github.com/JuliaLang/julia/blob/fbb3e1175d52abec0ff4ca83d8c9e126d9f8a06b/base/abstractarray.jl#L42

Issue #144, opened 2024-10-06
Custom adjacency graph from bipartite graph


gdalle:
Useful for bicoloring to avoid allocations

Related:

- #23

amontoison:
i think that we can close this issue.

Issue #122, opened 2024-09-27
Postprocessing for symmetric coloring in bicoloring


gdalle:
We need some way to detect useless colors

amontoison:
@gdalle 
```julia
using SparseMatrixColorings

###################################################################
# Rectangle matrix
###################################################################

m = 20
n = 30
p = 2
J = zeros(m,n)
for i = 1:p
  J[i,:] .= rand()
end
for i = m-p+1:m
  J[i,:] .= rand()
end
for j = 1:p
  J[:,j] .= rand()
end
for j = n-p+1:n
  J[:,j] .= rand()
end

J = sparse(J)
H = [0*I J'; J 0*I]

###################################################################
# rectangle -- star bicoloring
###################################################################

problem = ColoringProblem(; structure=:nonsymmetric,
                            partition=:bidirectional)

order = NaturalOrder()

algo = GreedyColoringAlgorithm(order;
                               decompression=:direct)

result = coloring(J, problem, algo)

rcolors = row_colors(result)
ccolors = column_colors(result)
num_colors = ncolors(result)

Br, Bc = compress(J, result)

# H2[i,j] is the color of the column j of H.
#
# Because of the shape of H:
# - If the column j of H represents a column of J (1 ≤ j ≤ n),
# H2[i,j] stores the color of the row `i-n` of J such that J[i-n,j] != 0.
# - If the column j of H represents a row of J (n+1 ≤ j ≤ n+m),
# H2[i,j] stores the color of the column `i` of J such that J[j-n,i] != 0.
H2 = copy(H)
for j in 1:n+m
  for index in H2.colptr[j] : H2.colptr[j+1]-1
    i = H2.rowval[index]
    if i > n
      H2.nzval[index] = rcolors[i-n]
    else
      H2.nzval[index] = ccolors[i]
    end
  end
end
H2 = Int.(H2)

# `checker` is a vector of size `num_colors` that will store, for a given column `j` of H,
# the number of times each color is used in `rcolors` or `ccolors`.
#
# `checker[c]` stores the number of times that color `c` appears in column `j` of H2.
#
# If each non-zero entry in column `j` of H2 has a different color
# (checker[c] <= 1 for all c), it means we can use a neutral color 0.
# This implies that all non-zero entries in the column `j` of H can be recovered from the rows `i` of H where H[i, j] != 0.
checker = Vector{Int}(undef, num_colors)

# List of columns of H that be colored with the neutral colors 0.
[...]

PR #146, opened 2024-10-07
Working prototype for bicoloring (coloring + decompression)

gdalle:
> [!NOTE]
> This PR is not about performance, it's about getting stuff to work and adding the necessary tests. We will worry about performance later.

- Fix a bug in `AdjacencyFromBipartiteGraph` (introduced in #144) where the neighbors didn't have the correct offset in the block structure.
- Define a type `BicoloringResult` which contains colors for both columns and rows, as well as a pointer to a `StarSetColoringResult`/`TreeSetColoringResult` associated with the big matrix `[0 A'; A 0]`.
- To go from column colors of `[0 A'; A 0]` to row and column colors of `A`, renumber from `1` on both sides. Example: if `size(A) = (3, 2)` and `color = [1, 2, 4, 3, 2]`, then `column_color = [1, 2, 3]` and `row_color = [2, 1]`. We lose the correspondence with the symmetric matrix, but we gain consistency within each compressed matrix `Br` and `Bc`. 
- Implement `compress`, `decompress` and `decompress!` to work with bidirectional colorings, using the row compression `Br` and the column compression `Bc`.
- For `decompress!`, the implementation is very suboptimal and should be improved in further PRs. Right now it recreates `[0 A'; A 0]`, decompresses into that big matrix and then keeps the bottom left corner. The upside is that it is entirely based on existing code for symmetric decompression.
- Add tests on random instances. It's important to use the random order here to avoid coloring all the columns first and then all the rows.

### DIFF
--- a/src/decompression.jl
+++ b/src/decompression.jl
@@ -571,10 +571,10 @@ function decompress!(
     B = zeros(T, n + m, length(unique(symmetric_color)))
     for c in axes(B, 2)
         if haskey(col_color_ind, c)  # some columns were colored with c
-            B[:, c] .+= vcat(zeros(T, n), Bc[:, col_color_ind[c]])
+            B[(n + 1):(n + m), c] .+= @view Bc[:, col_color_ind[c]]
         end
         if haskey(row_color_ind, c)  # some rows were colored with c
-            B[:, c] .+= vcat(Br[row_color_ind[c], :], zeros(T, m))
+            B[1:n, c] .+= @view Br[row_color_ind[c], :]
         end
     end
     new_A = decompress(B, result.symmetric_result)
--- a/src/decompression.jl
+++ b/src/decompression.jl
@@ -76,7 +76,7 @@ end
 
 """
     decompress(B::AbstractMatrix, result::AbstractColoringResult)
-    decompress(Br::AbstractMatrix, Bc::AbstractMatrix, result::AbstractColoringResult)
+    decompress(Br::AbstractMatrix, Bc::AbstractMatrix, result::AbstractColoringResult{_,:bidirectional})
 
 Decompress `B` into a new matrix `A`, given a coloring `result` of the sparsity pattern of `A`.
 The in-place alternative is [`decompress!`](@ref).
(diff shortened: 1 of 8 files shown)


src/decompression.jl
```diff
@@ -535,3 +561,28 @@ function decompress!(
...
+    m, n = size(A)
+    T = Base.promote_eltype(Br, Bc)
+    symmetric_color = column_colors(result.symmetric_result)
+    _, col_color_ind = remap_colors(symmetric_color[1:n])
+    _, row_color_ind = remap_colors(symmetric_color[(n + 1):(n + m)])
+    B = Matrix{T}(undef, n + m, maximum(symmetric_color))
+    fill!(B, zero(T))
+    for c in axes(B, 2)
+        if haskey(col_color_ind, c)  # some columns were colored with c
+            @views copyto!(B[(n + 1):(n + m), c], Bc[:, col_color_ind[c]])
```
amontoison, line 579: @gdalle Tu confirmes que `col_color_ind[c]` est un mapping entre les couleurs des colonnes 1 à n de la matrice par bloc 2x2 et les colonnes de `Bc`, et que `row_color_ind[c]` est un mapping entre les couleurs des colonnes n+1 à n+m de la matrice par bloc 2x2 et les lignes de `Br` ?

amontoison (approved):

amontoison:
Did you tested on the "rectangle" sparsity pattern?

src/decompression.jl
gdalle: exactly

gdalle:
> Did you tested on the "rectangle" sparsity pattern?

No because we don't yet have the postprocessing to remove useless colors, we can keep track in https://github.com/gdalle/SparseMatrixColorings.jl/issues/122

gdalle:
No it's not, but for bicoloring it can help remove some colors

amontoison:
I just suggested the rectangle sparsity pattern because it's easy to check the result :wink:

gdalle:
https://github.com/gdalle/SparseMatrixColorings.jl/issues/147

PR #152, opened 2024-10-27
Bicoloring

gdalle:
**Features**

- Implement coloring, compression and decompression for `:bidirectional` partitions (so-called bicoloring).
  - It works both for `:direct` and `:substitution` decompression, by relying on the corresponding routines for symmetric matrices.
  - The `BicoloringResult` preallocates all necessary memory for allocation-free in-place decompression into `A::SparseMatrixCSC`. Decompression into `A::AbstractMatrix` still allocates because we take a submatrix of `A_and_Aᵀ::SparseMatrixCSC` without `view`.
- Implement visualization for bicoloring results.
- Export the function `ncolors` which counts distinct colors (mainly useful for bicoloring because one needs to sum rows and columns).

**Tests**

- Add correctness, type stability, allocation and visualization tests for bicoloring.
- Test for two different `decompression_eltype` values.

**Chores**

- Bump version to v0.4.9


src/decompression.jl
```diff
@@ -63,10 +60,25 @@ function compress(A, result::AbstractColoringResult{structure,:row}) where {stru
     return B
 end
 
+function compress(
+    A, result::AbstractColoringResult{structure,:bidirectional}
+) where {structure}
+    row_group = row_groups(result)
+    column_group = column_groups(result)
+    Br = stack(row_group; dims=1) do g
+        dropdims(sum(A[g, :]; dims=1); dims=1)
```
amontoison, line 69: Can we avoid the allocation of the slice `A[g, :]` here?

src/decompression.jl
```diff
@@ -63,10 +60,25 @@ function compress(A, result::AbstractColoringResult{structure,:row}) where {stru
...
+    end
+    Bc = stack(column_group; dims=2) do g
+        dropdims(sum(A[:, g]; dims=2); dims=2)
```
amontoison, line 72: Can we avoid the allocation of `A[:, g]`?

src/result.jl
```diff
@@ -55,6 +55,25 @@ Return a vector `group` such that for every color `c`, `group[c]` contains the i
...
 
+"""
+    ncolors(result::AbstractColoringResult)
+
+Return the number of different colors used to color the matrix.
+
+For bidirectional partitions, this number is the sum of the number of row colors and the number of column colors.
+"""
+function ncolors(res::AbstractColoringResult{structure,:column}) where {structure}
+    return maximum(column_colors(res))
```
amontoison, line 66: Can we not use `length(res.group)`?
It should be a little bit cheaper.

src/result.jl
```diff
@@ -528,3 +547,96 @@ function LinearSystemColoringResult(
...
+    m, n = size(A)
+    symmetric_color = column_colors(symmetric_result)
+    column_color, col_color_ind = remap_colors(symmetric_color[1:n])
+    row_color, row_color_ind = remap_colors(symmetric_color[(n + 1):(n + m)])
+    column_group = group_by_color(column_color)
+    row_group = group_by_color(row_color)
+    B = Matrix{R}(undef, n + m, maximum(column_colors(symmetric_result)))
+    large_colptr = copy(ag.S.colptr)
+    large_colptr[(n + 2):end] .= large_colptr[n + 1]  # last few columns are empty
+    large_rowval = ag.S.rowval[1:(end ÷ 2)]  # forget the second half of nonzeros
```
amontoison, line 627: Can we add use a view here?

amontoison (approved):

src/decompression.jl
```diff
@@ -632,3 +658,49 @@ function decompress!(
     end
     return A
 end
+
+## BicoloringResult
+
+function _reconstruct_B!(result::BicoloringResult, Br::AbstractMatrix, Bc::AbstractMatrix)
```
amontoison, line 664: Can you add more comments to describe this function?

src/interface.jl
```diff
@@ -49,7 +49,7 @@ Matrix coloring is often used in automatic differentiation, and here is the tran
 | -------- | ------- | --------------- | ---------------- | ----------- |
 | Jacobian | forward | `:nonsymmetric` | `:column`        | yes         |
 | Jacobian | reverse | `:nonsymmetric` | `:row`           | yes         |
-| Jacobian | mixed   | `:nonsymmetric` | `:bidirectional` | no          |
+| Jacobian | mixed   | `:nonsymmetric` | `:bidirectional` | yes         |
```
amontoison, line 52: :muscle: :muscle:

src/result.jl
```diff
@@ -528,3 +547,96 @@ function LinearSystemColoringResult(
         T_factorization,
     )
 end
+
+## Bicoloring result
+
+function remap_colors(color::Vector{Int})
```
amontoison, line 553: I understand what the goal of this function but we should add more comments for the code.

src/decompression.jl
gdalle: Yes, let's do it. Although note that `compress` is mostly for our own use in testing.

src/result.jl
gdalle: Sure

src/decompression.jl
gdalle: Of course

src/decompression.jl
gdalle: See answer above

src/decompression.jl
gdalle: Actually I tried but this completely changes the nature of the compressed object (which becomes a `SparseMatrixCSC` instead of a `Matrix`, even though it is dense. So I reverted it and I suggest we leave it that way.

gdalle: In particular it made some tests fail.

amontoison (approved):

PR #155, opened 2024-11-11
Add a function structurally_biorthogonal for star bicoloring

amontoison:
- Fix three typos (`symmetrix` and `matric`).
- Add a function `proper_length_bicoloring` to check the length of the two color vectors returned by a bicoloring.
- Add a function `structurally_biorthogonal` based on a "new" definition of star bicoloring.

I don't need to know if a star bicoloring on `A` was already defined like this before; it’s something we could easily derive from the definition of a symmetric star coloring on `[0 A'; A 0]`.


gdalle (approved):
well done!

PR #156, opened 2024-11-11
Improve how we build A_and_Aᵀ

### LINKED ISSUE #161: Lazy bicoloring adjacency graph
See discussion in #156

amontoison:
close #161 

If we want to store exactly `[0 A'; A 0]`, we can't use `symmetric_pattern` to build `Aᵀ`.
```julia
Aᵀ = if symmetric_pattern || A isa Union{Symmetric,Hermitian}
  A
else
  transpose(A)
end
```
We will have the wrong nonzeros in `A_and_Aᵀ` (but the correct sparsity pattern).

I'm wondering if we should store a different matrix than `b = A_and_Aᵀ` in `StarSetResult` or `TreeSetResult`.  
If I remember correctly, we want to store `B` to have the correct type for the function `decompress`, but in this case, it’s always a `SparseMatrixCSC`, whereas `A` could have a different type.

We probably only want the `AdjacencyGraph` of `A_and_Aᵀ` and not the matrix itself, and we could likely achieve this with a keyword argument for `AdjacencyGraph` or another constructor that takes two arguments (`A` and `Aᵀ`).

Off-topic: We can modify the function  `neighbors(g::AdjacencyGraph, v::Integer)` to not use `filter` in the case of bicoloring because we know that we don't have diagonal coefficients.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -234,18 +234,72 @@ function coloring(
     decompression_eltype::Type{R}=Float64,
     symmetric_pattern::Bool=false,
 ) where {decompression,R}
+
+    # Build an AdjacencyGraph for the following matrix:
+    # [ 0  Aᵀ ]
+    # [ A  0  ]
     m, n = size(A)
-    T = eltype(A)
-    Aᵀ = if symmetric_pattern || A isa Union{Symmetric,Hermitian}
-        A
+    p = m + n
+    S = A isa SparseMatrixCSC ? A : SparseMatrixCSC(A)
+    nnzS = nnz(S)
+    rowval = Vector{Int}(undef, 2 * nnzS)
+    colptr = zeros(Int, p + 1)
+
+    # Update rowval and colptr for the block A
+    for i in 1:nnzS
+        rowval[i] = S.rowval[i] + n
+    end
+    for j in 1:n
+        colptr[j] = S.colptr[j]
+    end
+
+    # Update rowval and colptr for the block Aᵀ
+    if symmetric_pattern
+        # We use the sparsity pattern of A for Aᵀ
+        for i in 1:nnzS
+            rowval[nnzS + i] = S.rowval[i]
+        end
+        # m and n are identical because symmetric_pattern is true
+        for j in 1:m
+            colptr[n + j] = nnzS + S.colptr[j]
+        end
+        colptr[p + 1] = 2 * nnzS + 1
     else
-        transpose(A)
-    end  # TODO: fuse with next step?
-    A_and_Aᵀ = [
-        spzeros(T, n, n) SparseMatrixCSC(Aᵀ)
-        SparseMatrixCSC(A) spzeros(T, m, m)
-    ]  # TODO: slow
-    ag = AdjacencyGraph(A_and_Aᵀ; has_diagonal=false)
+        # We need to determine the sparsity pattern of Aᵀ
+        # We adapt the code of transpose(SparsityPatternCSC) in graph.jl
+        for k in 1:nnzS
+            i = S.rowval[k]
+            colptr[n + i] += 1
+        end
+
+        counter = 1
+        for col in (n + 1):p
+            nnz_col = colptr[col]
+            colptr[col] = counter
+            counter += nnz_col
+        end
+
+        for j in 1:n
+            for index in S.colptr[j]:(S.colptr[j + 1] - 1)
+                i = S.rowval[index]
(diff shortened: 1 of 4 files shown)


gdalle:
> We probably only want the AdjacencyGraph of A_and_Aᵀ and not the matrix itself, and we could likely achieve this with a keyword argument for AdjacencyGraph or another constructor that takes two arguments (A and Aᵀ).

You're right, in fact that's how I did it initially in #152 (check out the first commit, which is itself a sub-PR). I removed it afterwards because it felt simpler, but looking at the code again we probably never need to actually materialize `A_and_Aᵀ`. The lazy graph approach would likely work. I have opened #161 to keep track.

src/interface.jl
```diff
@@ -232,15 +232,22 @@ function coloring(
...
-        transpose(A)
-    end  # TODO: fuse with next step?
-    A_and_Aᵀ = [
-        spzeros(T, n, n) SparseMatrixCSC(Aᵀ)
-        SparseMatrixCSC(A) spzeros(T, m, m)
-    ]  # TODO: slow
+
+    # build A_and_Aᵀ
+    sparse_A = A isa SparseMatrixCSC ? A : SparseMatrixCSC(A)
+    nnzA = nnz(A)
```
gdalle, line 238:
```suggestion
    nnzA = nnz(sparse_A)
```
I think that's why the tests fail

src/interface.jl
```diff
@@ -232,15 +232,22 @@ function coloring(
...
+    colptr = Vector{Int}(undef, m + n + 1)
+    rowval = Vector{Int}(undef, 2 * nnzA)
+    nzval = Vector{T}(undef, 2 * nnzA)
+    colptr[1:(n + 1)] .= sparse_A.colptr
+    rowval[1:nnzA] .= sparse_A.rowval .+ n
+    nzval[1:nnzA] .= sparse_A.nzval
+    sparse_Aᵀ = SparseMatrixCSC(sparse_A')
```
gdalle, line 245:
```suggestion
    sparse_Aᵀ = SparseMatrixCSC(transpose(sparse_A))
```

gdalle (changes requested):

amontoison:
@gdalle 
I updated this old PR to directly build the augmented adjacency matrix in the format `SparsityPatternCSC`.
I also updated the definition such that `SparsityPatternCSC <: AbstractMatrix{Bool}`.
`StarSetColoringResult` and `TreeSetColoringResult` were not happy that `A_and_Aᵀ` was not an `AbstractMatrix`.

We should check with some benchmarks that we still target the optimized star / acyclic decompression for `SparseMatrixCSC`.

src/interface.jl
```diff
@@ -234,18 +234,72 @@ function coloring(
...
+            for index in S.colptr[j]:(S.colptr[j + 1] - 1)
+                i = S.rowval[index]
+                pos = colptr[n + i]
+                rowval[nnzS + pos] = j
+                colptr[n + i] += 1
+            end
+        end
+
+        colptr[p + 1] = nnzS + counter
+        @assert colptr[p + 1] == 2 * nnzS + 1
```
amontoison, line 292:
```suggestion
```

gdalle (changes requested):
Can we reduce code duplication by using the same code for `transpose` and `bidirectional_pattern`? For instance defining a subroutine like

```julia
transpose_pattern!(A, colptr, rowval)
```

which could then also be applied on `colptr = view(...)` and `rowval = view(...)` in the case of the bigger matrix (and followed by a vectorized addition to account for the shift)

amontoison:
> Can we reduce code duplication by using the same code for `transpose` and `bidirectional_pattern`? For instance defining a subroutine like
> 
> ```julia
> transpose_pattern!(A, colptr, rowval)
> ```
> 
> which could then also be applied on `colptr = view(...)` and `rowval = view(...)` in the case of the bigger matrix (and followed by a vectorized addition to account for the shift)

Yes, we could reduce code duplication by implementing a subroutine like `transpose_pattern!`.
However, the gain would only be around 20–30 lines of code (including comments and spacing).

That said, introducing the offset logic for rows, columns, and totals might slightly impact readability.
Personally, I don't think it's worth spending time on this right now.

I suggest we merge the PR as it is.
If you feel strongly about reducing duplication, we could open an issue to revisit this improvement later.

gdalle (approved):
i don't fully understand the code but I trust the tests I added

PR #157, opened 2024-11-11
Add compressed visualization

gdalle:
- [x] Change `show_colors` so that it includes the visualization of the compressed matrix. It now returns a tuple `(A_img, B_img)`  or `(A_img, Br_img, Bc_img)`.
- [x] Demonstrate compressed visualization in the docs, on a column coloring and on a bidirectional coloring.
- [x] Test the tail of the tuple (right now only `A_img` is tested for consistency).


gdalle:
> Can you also support the neutral color 0?

We could but at the moment there are lots of spots where we implicitly assume the colors start at `1`. So I don't think it would make much sense to change this only in visualization. I have opened #158 to keep track of this more general overhaul.

amontoison (approved):

PR #171, opened 2025-02-12
Optimize star coloring and its postprocessing

amontoison:
- Preallocate the number of entries (i.e., the number of edges) in the dictionary `star`.
- Add a vector of integers `nb_spokes` to count the number of spokes for each star during the star coloring.
- Use `nb_spokes` to know the size of each vector in `spokes` when creating a `StarSet`.
- Restore the negative hubs in the trivial stars so that we can treat them after the "normal stars" during postprocessing.
- Improve handling of trivial stars during postprocessing by checking if a vertex is already a hub in a different star.
- Improve `_update_stars!` to minimize duplicated access in the dictionary `star`.


amontoison:
@gdalle The failures are related to the decompression of bicoloring.
It should be related to the neutral colors that I optimized.
I think that we neutralized some colors now with the random tests.

src/coloring.jl
```diff
@@ -543,10 +560,32 @@ function postprocess!(
...
+        end
+
+        # Process the trivial stars (if any)
+        if nb_trivial_star > 0
+            for s in eachindex(hub)
+                j = hub[s]
+                if j < 0
+                    i = spokes[s][1]
+                    j = abs(j)
+                    if !color_used[color[i]] && !color_used[color[j]]
```
gdalle, line 583: @amontoison I think the problem is here. During decompression, we always recover coefficients based on the color of the star's hub. With this change, if `color_used[color[i]]`, we're not assigning any color to `j` at all. Either we revert this or we adapt `symmetric_coefficient` accordingly. I'll try adapting.

gdalle: The hub swapping seemed to solve it

gdalle (approved):

src/coloring.jl
amontoison: Good catch Guillaume!

We probably need to a similar modification with the trivial trees (when we will handle them).

amontoison:
The gain in terms of storage is between `24%` and `67%` for the star coloring.

PR #174, opened 2025-02-12
Optimize ordering for bipartite graphs

gdalle:
More clever computation of distance-two degrees

### DIFF
--- a/src/order.jl
+++ b/src/order.jl
@@ -219,15 +219,28 @@ function vertices(
     g::BipartiteGraph, ::Val{side}, ::DynamicDegreeBasedOrder{degtype,direction}
 ) where {side,degtype,direction}
     other_side = 3 - side
-    if degree_increasing(; degtype, direction)
-        degrees = zeros(Int, nb_vertices(g, Val(side)))
-    else
-        degrees = [degree_dist2(g, Val(side), v) for v in vertices(g, Val(side))]  # TODO: optimize
+    # compute dist-2 degrees in an optimized way
+    n = nb_vertices(g, Val(side))
+    degrees_dist2 = zeros(Int, n)
+    dist2_neighbor = falses(n)
+    for v in vertices(g, Val(side))
+        fill!(dist2_neighbor, false)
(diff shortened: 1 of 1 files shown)


src/order.jl
```diff
@@ -219,15 +219,28 @@ function vertices(
...
+                dist2_neighbor[w2] = true
+            end
+        end
+        degrees_dist2[v] = sum(dist2_neighbor)
+    end
     if degree_increasing(; degtype, direction)
-        degrees = zeros(Int, nb_vertices(g, Val(side)))
+        degrees = zeros(Int, n)
     else
-        degrees = [degree_dist2(g, Val(side), v) for v in vertices(g, Val(side))]  # TODO: optimize
```
amontoison, line 225: Do we still use the function `degree_dist2` somewhere?

src/order.jl
```diff
@@ -219,15 +219,28 @@ function vertices(
...
+    degrees_dist2 = zeros(Int, n)
+    dist2_neighbor = falses(n)
+    for v in vertices(g, Val(side))
+        fill!(dist2_neighbor, false)
+        for w1 in neighbors(g, Val(side), v)
+            for w2 in neighbors(g, Val(other_side), w1)
+                dist2_neighbor[w2] = true
+            end
+        end
+        degrees_dist2[v] = sum(dist2_neighbor)
```
amontoison, line 233: I suggest to directly update `degrees_dist2[v]` after line 230 with a `degrees_dist2[v] += 1`.

amontoison: Update: I am wrong if we can visit `w2` multiple times (`dist2_neighbor[w2] = true`).
If it's the case, forget my comment.

gdalle: yes we can visit multiple times

src/order.jl
gdalle: only in the test suite, but it doesn't hurt keeping it (coverage hasn't dropped)

amontoison (approved):
I added a few comments but you should be able to merge after.

PR #176, opened 2025-02-12
Reduce the dynamic allocations in acyclic coloring

amontoison:
I wanted to use the function `sizehint!` in `DataStructures.jl` but it's incomplete:
https://github.com/JuliaCollections/DataStructures.jl/blob/master/src/disjoint_set.jl#L182-L186

We should do our own structure for the `forest` one day.


amontoison:
@gdalle We gain between `11%` and `33%` of storage for the acyclic coloring with this modification.

gdalle:
but it seems we lose in runtime?

gdalle:
sorry I was reading the table wrong, this is great!

gdalle (approved):

PR #177, opened 2025-02-12
Add rectangle example

gdalle:


test/small.jl
```diff
@@ -85,3 +86,33 @@ end;
...
+    result = coloring(
+        A, problem, GreedyColoringAlgorithm{:direct}(order; postprocessing=false)
+    )
+    @test ncolors(result) == 6  # two more than necessary
+    result = coloring(
+        A, problem, GreedyColoringAlgorithm{:direct}(order; postprocessing=true)
+    )
+    @test ncolors(result) == 4  # optimal number
+
+    # TODO: find example where subsitution is strictly better
```
amontoison, line 109: It's the case with the arrowhead matrix.

amontoison (approved):

test/small.jl
gdalle: ok I'll add it real quick

amontoison: I have these examples if you want:
```julia
if triangle
  m = 3
  n = 3
  J = [1 1 0; 0 1 1; 1 0 1]
  J = sparse(J)
  H = [0*I J'; J 0*I]
end
```
```julia
if rectangle
  m = 6
  n = 6
  p = 1
  J = zeros(m,n)
  for i = 1:p
    J[i,:] .= rand()
  end
  for i = m-p+1:m
    J[i,:] .= rand()
  end
  for j = 1:p
    J[:,j] .= rand()
  end
  for j = n-p+1:n
    J[:,j] .= rand()
  end
end
```
```julia
if arrowhead
  m = 50
  n = 50
  p = 2
  J = zeros(m,n)
  for i = m-p:m
    J[i,:] .= rand()
  end
  for j = n-p:n
    J[:,j] .= rand()
  end
  for i = 1:m
    for j = 1:n
      if abs(i - j) ≤ p
        J[i,j] = rand()
      end
    end
  end
end
```

gdalle: it seems to work even with the simplest one

amontoison: The triangle is great for the code coverage because we can only have trivial stars and trees after a star and acyclic coloring, respectively.

PR #179, opened 2025-02-13
Remove vertices_by_tree for the acyclic coloring

amontoison:
This PR removes `vertices_by_tree` from the `TreeSet`.
This `Vector{Vector{Int}}` contains the list of vertices for each tree in the forest.  
It was previously used to reset the buffer during the decompression of the acyclic coloring.  
However, we can easily recover the vertices of a tree from the list of edges of the related tree (`reverse_bfs_orders`).
Since all edges of a tree are stored in the format `(leaf, parent)`, we can reset the buffer for all "leaves" plus the parent of the last edge, which corresponds to the root of the tree.

Other minor modifications:
- Specify the type of some arrays (`spokes`, `trees`, `reverse_bfs_orders`) to potentially help the compiler. I remember that JET was not happy for the following allocations in Krylov.jl ([here](https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/krylov_solvers.jl#L2481)) because the compiler is not able to statically determine if the vector is empty or not.
- Use `sizehint!` for the dictionary `roots`.

### DIFF
--- a/src/coloring.jl
+++ b/src/coloring.jl
@@ -485,7 +485,7 @@ function TreeSet(forest::DisjointSets{Tuple{Int,Int}}, nvertices::Int)
     degrees = Vector{Int}(undef, nvertices)
 
     # reverse breadth first (BFS) traversal order for each tree in the forest
-    reverse_bfs_orders = Vector{Vector{Tuple{Int,Int}}}[Tuple{Int,Int}[] for i in 1:ntrees]
+    reverse_bfs_orders = Vector{Tuple{Int,Int}}[Tuple{Int,Int}[] for i in 1:ntrees]
 
     # nvmax is the number of vertices of the biggest tree in the forest
     nvmax = 0
--- a/src/coloring.jl
+++ b/src/coloring.jl
@@ -148,7 +148,7 @@ end
 
 function StarSet(star::Dict{Tuple{Int,Int},Int}, hub::Vector{Int}, nb_spokes::Vector{Int})
     # Create a list of spokes for each star, preallocating their sizes based on nb_spokes
-    spokes = Vector{Int}[Vector{Int}(undef, ns) for ns in nb_spokes]
+    spokes = [Vector{Int}(undef, ns) for ns in nb_spokes]
 
     # Reuse nb_spokes as counters to track the current index while filling the spokes
     fill!(nb_spokes, 0)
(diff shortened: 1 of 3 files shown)


amontoison:
Small gain in storage for the acyclic coloring -- between `4%` and `9%`.

src/coloring.jl
```diff
@@ -148,7 +148,7 @@ end
 
 function StarSet(star::Dict{Tuple{Int,Int},Int}, hub::Vector{Int}, nb_spokes::Vector{Int})
     # Create a list of spokes for each star, preallocating their sizes based on nb_spokes
-    spokes = [Vector{Int}(undef, ns) for ns in nb_spokes]
+    spokes = Vector{Int}[Vector{Int}(undef, ns) for ns in nb_spokes]
```
gdalle, line 151:
```suggestion
    spokes = [Vector{Int}(undef, ns) for ns in nb_spokes]
```
the additional annotation is not necessary

src/coloring.jl
```diff
@@ -445,16 +444,16 @@ function TreeSet(forest::DisjointSets{Tuple{Int,Int}}, nvertices::Int)
 
     # dictionary that maps a tree's root to the index of the tree
     roots = Dict{Int,Int}()
+    sizehint!(roots, ntrees)
 
     # vector of dictionaries where each dictionary stores the neighbors of each vertex in a tree
-    trees = [Dict{Int,Vector{Int}}() for i in 1:ntrees]
+    trees = Dict{Int,Vector{Int}}[Dict{Int,Vector{Int}}() for i in 1:ntrees]
```
gdalle, line 450:
```suggestion
    trees = [Dict{Int,Vector{Int}}() for i in 1:ntrees]
```

src/coloring.jl
```diff
@@ -485,14 +484,15 @@ function TreeSet(forest::DisjointSets{Tuple{Int,Int}}, nvertices::Int)
     # degrees is a vector of integers that stores the degree of each vertex in a tree
     degrees = Vector{Int}(undef, nvertices)
 
-    # list of vertices for each tree in the forest
-    vertices_by_tree = [collect(keys(trees[i])) for i in 1:ntrees]
-
     # reverse breadth first (BFS) traversal order for each tree in the forest
-    reverse_bfs_orders = [Tuple{Int,Int}[] for i in 1:ntrees]
+    reverse_bfs_orders = Vector{Tuple{Int,Int}}[Tuple{Int,Int}[] for i in 1:ntrees]
```
gdalle, line 488:
```suggestion
    reverse_bfs_orders = [Tuple{Int,Int}[] for i in 1:ntrees]
```

gdalle (approved):
lgtm apart from the type annotations, which are not helpful

PR #180, opened 2025-02-13
Detect stars in the forest for acyclic coloring

### LINKED ISSUE #178: Try to detect the stars during the acyclic coloring
We will have a more efficient postprocessing if we know that some trees are stars.

amontoison:
close #178, #181
Add a vector of booleans `is_star` to the `TreeSet` so that we can enable more efficient post-processing for acyclic coloring by determining if a tree is a star (trivial or non-trivial).

### DIFF
--- a/src/coloring.jl
+++ b/src/coloring.jl
@@ -435,6 +435,7 @@ $TYPEDFIELDS
 """
 struct TreeSet
     reverse_bfs_orders::Vector{Vector{Tuple{Int,Int}}}
+    is_star::Vector{Bool}
 end
 
 function TreeSet(forest::DisjointSets{Tuple{Int,Int}}, nvertices::Int)
@@ -499,8 +500,18 @@ function TreeSet(forest::DisjointSets{Tuple{Int,Int}}, nvertices::Int)
     # Create a queue with a fixed size nvmax
     queue = Vector{Int}(undef, nvmax)
 
+    # Specify if each tree in the forest is a star,
+    # meaning that one vertex is directly connected to all other vertices in the tree
+    is_star = Vector{Bool}(undef, ntrees)
+
     for k in 1:ntrees
         tree = trees[k]
+
+        # Boolean indicating whether the current tree is a star (a single central vertex connected to all others)
+        bool_star = true
+
+        # Candidate hub vertex if the current tree is a star
+        virtual_hub = 0
 
         # Initialize the queue to store the leaves
         queue_start = 1
@@ -527,9 +538,23 @@ function TreeSet(forest::DisjointSets{Tuple{Int,Int}}, nvertices::Int)
             degrees[leaf] = 0
 
             for neighbor in tree[leaf]
+                # Check if neighbor is the parent of the leaf or if it was a child before the tree was pruned
                 if degrees[neighbor] != 0
                     # (leaf, neighbor) represents the next edge to visit during decompression
                     push!(reverse_bfs_orders[k], (leaf, neighbor))
+
+                    if bool_star
+                        # Initialize the potential hub of the star with the first parent of a leaf
+                        if virtual_hub == 0
+                            virtual_hub = neighbor
+                        else
+                            # Verify if the tree still qualifies as a star
+                            # If we find leaves with different parents, then it can't be a star
+                            if virtual_hub != neighbor
+                                bool_star = false
+                            end
+                        end
+                    end
 
                     # reduce the degree of the neighbor
                     degrees[neighbor] -= 1
@@ -542,9 +567,12 @@ function TreeSet(forest::DisjointSets{Tuple{Int,Int}}, nvertices::Int)
                 end
             end
         end
+
+        # Specify if the tree is a star or not
+        is_star[k] = bool_star
     end
 
-    return TreeSet(reverse_bfs_orders)
+    return TreeSet(reverse_bfs_orders, is_star)
 end
 
 ## Postprocessing, mirrors decompression code
@@ -605,17 +633,23 @@ function postprocess!(
         end
     else
         # only the colors of non-leaf vertices are used
-        (; reverse_bfs_orders) = star_or_tree_set
+        (; reverse_bfs_orders, is_star) = star_or_tree_set
         nb_trivial_trees = 0
 
         # Iterate through all non-trivial trees
         for k in eachindex(reverse_bfs_orders)
             reverse_bfs_order = reverse_bfs_orders[k]
-            # Check if we have more than one edge in the tree
+            # Check if we have more than one edge in the tree (non-trivial tree)
             if length(reverse_bfs_order) > 1
-                # TODO: Optimize by avoiding iteration over all edges
-                # Only one edge is needed if we know if it is a normal tree or a star
-                for (i, j) in reverse_bfs_order
+                # Determine if the tree is a star
+                if is_star[k]
+                    # It is a non-trivial star and only the color of the hub is needed
+                    (_, hub) = reverse_bfs_order[1]
+                    color_used[color[hub]] = true
+                else
+                    # It is not a star and both colors are needed during the decompression
+                    (i, j) = reverse_bfs_order[1]
+                    color_used[color[i]] = true
                     color_used[color[j]] = true
                 end
             else
--- a/src/decompression.jl
+++ b/src/decompression.jl
@@ -567,6 +567,7 @@ function decompress!(
         for (i, j) in reverse_bfs_orders[k]
             val = B[i, color[j]] - buffer_right_type[i]
             buffer_right_type[j] = buffer_right_type[j] + val
+
             if in_triangle(i, j, uplo)
                 A[i, j] = val
             end
--- a/test/allocations.jl
+++ b/test/allocations.jl
@@ -19,7 +19,7 @@ end
 
 @testset "Distance-2 coloring" begin
     test_noallocs_distance2_coloring(1000)
-end;
+end
 
 function test_noallocs_sparse_decompression(
     n::Integer; structure::Symbol, partition::Symbol, decompression::Symbol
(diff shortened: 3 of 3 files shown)


src/coloring.jl
```diff
@@ -599,17 +625,23 @@ function postprocess!(
...
-            # Check if we have more than one edge in the tree
+            # Check if we have more than one edge in the tree (non-trivial tree)
             if length(reverse_bfs_order) > 1
-                # TODO: Optimize by avoiding iteration over all edges
-                # Only one edge is needed if we know if it is a normal tree or a star
-                for (i, j) in reverse_bfs_order
+                # Determine if the tree is a star
+                if is_star[k]
+                    # It is a non-trivial star and only the color of the hub is needed
+                    (_, hub) = reverse_bfs_order[end]
```
amontoison, line 639: We could also use `reverse_bfs_order[1]`, it could be a little bit more efficient when we compute the address.
It's a micro-optimization...
@gdalle What is most relevant for readability?
I think both are fine.

amontoison, line 639:
```suggestion
                    (_, hub) = reverse_bfs_order[1]
```

amontoison:
I merged #184 in this branch, it only contained  4 modifications.

src/coloring.jl
```diff
@@ -602,17 +628,23 @@ function postprocess!(
...
```
amontoison, line 642:
```suggestion
                    (_, hub) = reverse_bfs_order[1]
```

src/coloring.jl
```diff
@@ -529,6 +540,18 @@ function TreeSet(forest::DisjointSets{Tuple{Int,Int}}, nvertices::Int)
...
 
+                    if bool_star
+                        # Initialize the potential hub of the star with the first parent of a leaf
+                        if virtual_hub == 0
+                            virtual_hub = neighbor
+                        else
+                            # Verify if the tree still qualifies as a star
+                            if virtual_hub != neighbor
+                                bool_star = false
+                            end
```
gdalle, line 551: I don't understand this logic

src/decompression.jl
```diff
@@ -616,53 +632,98 @@ function decompress!(
...
-
-                # uplo = :L or uplo = :F
-                # A[j,i] is stored at index_ji = (A.colptr[i+1] - offset_L) in A.nzval
-                if uplo != :U
-                    nzind = A_colptr[i + 1] - lower_triangle_offsets[counter]
-                    nzA[nzind] = val
+                #! format: on
+            end
+        else
+            # The tree is a star (trivial or non-trivial)
```
gdalle, line 687: Can we reduce code duplication here?

src/coloring.jl
amontoison: We define as "virtual hub the first parent of an edge that we encounter.
We then check that the parent of all other edges in the tree is "virtual hub".
If it's true, we have a star.

src/decompression.jl
amontoison: I think that we should define an auxiliary function for that

src/coloring.jl
amontoison: I added two comments in the code. Please don't hesitate to say if it is still unclear.

src/decompression.jl
amontoison: I added a function `compute_tree_value` that determine the value of `A[i,j]` / `A[j,i]`.
It takes into account if the tree is a star or not with the help of multiple dispatch for a faster recovery.

src/decompression.jl
```diff
@@ -514,16 +514,29 @@ end
...
 )
-    (; ag, color, reverse_bfs_orders, buffer) = result
+    (; ag, color, reverse_bfs_orders, is_star, buffer) = result
     (; S) = ag
     uplo == :F && check_same_pattern(A, S)
     R = eltype(A)
     fill!(A, zero(R))
 
-    if eltype(buffer) == R
+    if eltype(buffer) == R || isempty(buffer)
```
gdalle, line 539: Emptiness can't be inferred statically, as a result there may be a type instability here. I suggest we get rid of this small buffer optimization
```suggestion
    if eltype(buffer) == R
```

src/decompression.jl
```diff
@@ -540,17 +553,23 @@ function decompress!(
 
     # Recover the off-diagonal coefficients of A
     for k in eachindex(reverse_bfs_orders)
-        # Reset the buffer to zero for all vertices in a tree (except the root)
-        for (vertex, _) in reverse_bfs_orders[k]
-            buffer_right_type[vertex] = zero(R)
+        is_star_k = is_star[k]
+        val_is_star_k = Val(is_star_k)
```
gdalle, line 557: This is type-unstable, we can do something simpler

src/result.jl
```diff
@@ -361,14 +362,16 @@ function TreeSetColoringResult(
 
     # buffer holds the sum of edge values for subtrees in a tree.
     # For each vertex i, buffer[i] is the sum of edge values in the subtree rooted at i.
-    buffer = Vector{R}(undef, nvertices)
+    # Note that we don't need a buffer is all trees are stars.
+    buffer = all(is_star) ? R[] : Vector{R}(undef, nvertices)
```
gdalle, line 366:
```suggestion
    buffer = Vector{R}(undef, nvertices)
```
Because the check `isempty(buffer)` is not statically knowable

src/decompression.jl
```diff
@@ -616,18 +636,23 @@ function decompress!(
...
```
gdalle, line 640: This is type-unstable too

src/decompression.jl
```diff
@@ -514,16 +514,29 @@ end
...
+    val = B[i, color[j]] - buffer[i]
+    buffer[j] = buffer[j] + val
+    return val
+end
+
+function compute_tree_value(is_star::Val{true}, B::AbstractMatrix, i::Integer, j::Integer, color::AbstractVector{<:Integer}, buffer::AbstractVector{<:Real})
+    # The tree is a star (trivial or non-trivial)
+    val = B[i, color[j]]
+    return val
+end
```
gdalle, line 528:
```suggestion
function compute_tree_value(is_star::Bool, B::AbstractMatrix, i::Integer, j::Integer, color::AbstractVector{<:Integer}, buffer::AbstractVector{<:Real})
    if !is_star
        # The tree is not a star
        val = B[i, color[j]] - buffer[i]
        buffer[j] = buffer[j] + val
        return val
    else
        # The tree is a star (trivial or non-trivial)
        val = B[i, color[j]]
        return val
    end
end
```

src/decompression.jl
gdalle:
```suggestion
```

src/decompression.jl
```diff
@@ -540,17 +553,23 @@ function decompress!(
...
+            buffer_right_type[root] = zero(R)
         end
-        # Reset the buffer to zero for the root vertex
-        (_, root) = reverse_bfs_orders[k][end]
-        buffer_right_type[root] = zero(R)
 
         for (i, j) in reverse_bfs_orders[k]
-            val = B[i, color[j]] - buffer_right_type[i]
-            buffer_right_type[j] = buffer_right_type[j] + val
+            val = compute_tree_value(val_is_star_k, B, i, j, color, buffer_right_type)
```
gdalle, line 571:
```suggestion
            val = compute_tree_value(is_star_k, B, i, j, color, buffer_right_type)
```

src/decompression.jl
```diff
@@ -583,7 +603,7 @@ function decompress!(
     nzA = nonzeros(A)
     uplo == :F && check_same_pattern(A, S)
 
-    if eltype(buffer) == R
+    if eltype(buffer) == R || isempty(buffer)
```
gdalle, line 606:
```suggestion
    if eltype(buffer) == R
```

src/decompression.jl
gdalle:
```suggestion
```

src/decompression.jl
```diff
@@ -616,18 +636,23 @@ function decompress!(
...
         end
-        # Reset the buffer to zero for the root vertex
-        (_, root) = reverse_bfs_orders[k][end]
-        buffer_right_type[root] = zero(R)
 
         for (i, j) in reverse_bfs_orders[k]
             counter += 1
-            val = B[i, color[j]] - buffer_right_type[i]
-            buffer_right_type[j] = buffer_right_type[j] + val
+            compute_tree_value(val_is_star_k, B, i, j, color, buffer_right_type)
```
gdalle, line 655:
```suggestion
            compute_tree_value(is_star_k, B, i, j, color, buffer_right_type)
```

test/allocations.jl
```diff
@@ -132,3 +132,21 @@ end;
...
+        result = coloring(A, problem, GreedyColoringAlgorithm{:substitution}())
+        @test isempty(result.buffer)
+        for T in (Float32, Float64)
+            C = rand(T) * T.(A)
+            B = compress(C, result)
+            bench_multiprecision = @be decompress!(C, B, result)
+            @test minimum(bench_multiprecision).allocs == 0
+        end
+    end
+end
```
gdalle, line 152: If we agree to remove the buffer optimization, this has to go too

src/decompression.jl
gdalle: The idea being that currently, the type of `buffer_right_type` is not statically known, which can be bad

gdalle (changes requested):

src/result.jl
amontoison: Can we use `Nothing` instead `R[]` for the buffer? It will solve the issue with `isempty` if we replace is with `isnothing`.

amontoison:
@gdalle I simplified the PR a lot, the boolean vector `is_star` is now only used for the post-processing for acyclic coloring / bicoloring.

gdalle (approved):

PR #186, opened 2025-02-14
Add an attribute has_loops in AdjacencyGraph

### LINKED ISSUE #182: Pass an option bicoloring to star_coloring and acyclic_coloring
When performing bicoloring, we already know that there are no coefficients on the diagonal of the symmetric matrix to which we apply the star or acyclic coloring.
We can use this information to avoid checking whether the symmetric matrix is zero on its diagonal during post-processing and to skip verifying if it needs to be recovered during decompression.

amontoison:
close #182
Take into account that the diagonal is zero in bicoloring and could be zero in symmetric colorings for a faster postprocessing, decompression and construction of the `TreeSetColoringResult`.

### DIFF
--- a/src/graph.jl
+++ b/src/graph.jl
@@ -121,10 +121,17 @@ struct AdjacencyGraph{T,has_diagonal}
     S::SparsityPatternCSC{T}
 end
 
-AdjacencyGraph(S::SparsityPatternCSC) = AdjacencyGraph{typeof(S),true}(S)
+function AdjacencyGraph(S::SparsityPatternCSC{T}; has_diagonal::Bool=true) where {T}
+    return AdjacencyGraph{T,has_diagonal}(S)
+end
 
-AdjacencyGraph(A::AbstractMatrix) = AdjacencyGraph(SparseMatrixCSC(A))
-AdjacencyGraph(A::SparseMatrixCSC) = AdjacencyGraph(SparsityPatternCSC(A))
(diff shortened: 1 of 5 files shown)


src/graph.jl
```diff
@@ -119,19 +119,41 @@ The adjacency graph of a symmetric matrix `A ∈ ℝ^{n × n}` is `G(A) = (V, E)
...
 function neighbors(g::AdjacencyGraph, v::Integer)
     S = pattern(g)
-    neighbors_with_loops = view(rowvals(S), nzrange(S, v))
-    return Iterators.filter(!=(v), neighbors_with_loops)  # TODO: optimize
+    neighbors_v = view(rowvals(S), nzrange(S, v))
+    if g.has_loops
+        return Iterators.filter(!=(v), neighbors_v)  # TODO: optimize
+    else
+        return neighbors_v
+    end
```
gdalle, line 156: This part worries me because the type of `Iterators.filter(...)` is different from the type of `neighbors_v`. To avoid type-instabilities, I suggest putting `has_loops` inside a type parameter

gdalle:
I just turned `has_loops` into a type parameter `has_diagonals`, which is statically set to false for bicoloring. This prevents us from exploiting this in the symmetric case in order to preserve type stability of the `neighbors` operator. Luckily, I think there won't be many Hessians with only zeros on the diagonal.

gdalle (approved):

Issue #143, opened 2024-10-06
Custom structure for disjoint trees


gdalle:
We need something better than `DataStructures.DisjointSets` to iteratively build a set of disjoint trees, keeping track of the BFS order

amontoison:
#190 will help because we will have our own structure `Forest`. We can probably update it to keep track of the BFS order.

amontoison:
I think we can't do better than our current structure `Forest`.

PR #190, opened 2025-02-14
Implement our own structure Forest

amontoison:
#143 
- Implement our own structure `Forest` with only the routines and attributes that we need for SMC.jl. We don't need anymore a vector with all edges stored (`revmap`), so it will be great for storage.
- `root_union!` returns the index of the root edge now. It was the root edge previously and we needed to recover the index with a dictionary. It's more efficient now.
- Drop the dependency on `DataStructures.jl`.
- Don't compress the forest independently; this is already handled during the creation of the TreeSet now. (see #166).


src/forest.jl
```diff
@@ -0,0 +1,52 @@
...
+    ntrees::T
+end
+
+function Forest{T}(n::Integer) where {T<:Integer}
+    counter = zero(T)
+    intmap = Dict{Tuple{T,T},T}()
+    sizehint!(intmap, n)
+    parents = collect(Base.OneTo(T(n)))
+    ranks = zeros(T, T(n))
+    ntrees = T(n)
```
amontoison, line 15:
```suggestion
    ntrees = zero(T)
```

src/forest.jl
```diff
@@ -0,0 +1,52 @@
...
+    ranks = zeros(T, T(n))
+    ntrees = zero(T)
+    return Forest{T}(counter, intmap, parents, ranks, ntrees)
+end
+
+function Base.push!(forest::Forest{T}, edge::Tuple{T,T}) where {T<:Integer}
+    forest.counter += 1
+    forest.intmap[edge] = forest.counter
+    forest.ntrees += one(T)
+    return edge
```
gdalle, line 23:
```suggestion
    return forest
```
typically, modifying methods return the object they're modifying

src/forest.jl
```diff
@@ -0,0 +1,52 @@
...
+end
+
+function _find_root!(parents::Vector{T}, index_edge::T) where {T<:Integer}
+    @inbounds p = parents[index_edge]
+    @inbounds if parents[p] != p
```
gdalle, line 28:
```suggestion
    p = parents[index_edge]
    if parents[p] != p
```
`@inbounds` is almost never a good idea unless we prove it is essential

src/forest.jl
```diff
@@ -0,0 +1,52 @@
...
+
+function find_root!(forest::Forest{T}, edge::Tuple{T,T}) where {T<:Integer}
+    return _find_root!(forest.parents, forest.intmap[edge])
+end
+
+function root_union!(forest::Forest{T}, index_edge1::T, index_edge2::T) where {T<:Integer}
+    parents = forest.parents
+    rks = forest.ranks
+    @inbounds rank1 = rks[index_edge1]
+    @inbounds rank2 = rks[index_edge2]
```
gdalle, line 42:
```suggestion
    rank1 = rks[index_edge1]
    rank2 = rks[index_edge2]
```

src/forest.jl
```diff
@@ -0,0 +1,52 @@
...
+
+    if rank1 < rank2
+        index_edge1, index_edge2 = index_edge2, index_edge1
+    elseif rank1 == rank2
+        rks[index_edge1] += one(T)
+    end
+    @inbounds parents[index_edge2] = index_edge1
```
gdalle, line 49:
```suggestion
    parents[index_edge2] = index_edge1
```

src/forest.jl
```diff
@@ -0,0 +1,52 @@
+mutable struct Forest{T<:Integer}
+    counter::T
+    intmap::Dict{Tuple{T,T},T}
+    parents::Vector{T}
+    ranks::Vector{T}
+    ntrees::T
+end
```
gdalle, line 7: can you explain what the fields are for?

gdalle (changes requested):
This looks really promising! In addition to the remarks below, would it be possible to add tests that specifically target the `Forest` structure, regardless from coloring?

src/forest.jl
amontoison: Can you explain why?
It's how was implemented `DisjointSet` in `DataStructures.jl`.
For performance reasons, I don't see why we should not use it.

gdalle: Older Julia libraries like DataStructures.jl had a much more liberal use of `@inbounds` than modern code bases. As the Julia compiler becomes smarter, it becomes less obvious where the macro actually speeds things up (see this [Discourse thread](https://discourse.julialang.org/t/when-does-inbounds-increase-performance/123827)).
But more importantly, misusing `@inbounds` can have very bad correctness implications or straight up crash the REPL (I've seen it with SparseDiffTools.jl for example). Here, your uses of `@inbounds` are unchecked, meaning that there is no guarantee for `index_edge` to be within the keys of `parents`.
I am very very very very strongly in favor of never using `@inbounds`. If you come up with an example where bounds checking alone takes more than 10% of the time, then yes we can reconsider (after adding proper checks), but it will also mean we have overcome every other complexity hurdle in our way. Until then, the risk is not worth the reward.

src/forest.jl
```diff
@@ -0,0 +1,69 @@
...
+"""
+$TYPEDEF
+
+Structure that provides fast union-find operations for constructing a forest during acyclic coloring and bicoloring.
+
+# Fields
+
+$TYPEDFIELDS
+"""
+$TYPEDFIELDS
```
amontoison, line 12:
```suggestion
```

amontoison:
@gdalle 
I added some unit tests for the structure `Forest` as you requested last week.

gdalle (approved):

PR #194, opened 2025-02-20
[bicoloring] Use Br and Bc directly for the decompression

amontoison:
Use a lazy representation of `Br_and_Br`.
The PR can be merged later but I wanted to check if it works to do the decompression like this.
I would like to describe the decompression of bicoloring with this approach is the paper where we directly use `Br` and `Bc`.


amontoison:
It seems that compress can return `Br` and `Bc` with a different type.
```
ERROR: LoadError: MethodError: no method matching SparseMatrixColorings.JoinCompressed(::Int64, ::Int64, ::Int64, ::Matrix{Int64}, ::SparseArrays.SparseMatrixCSC{Bool, Int64}, ::Dict{Int64, Int64}, ::Dict{Int64, Int64})
The type `SparseMatrixColorings.JoinCompressed` exists, but no method is defined for this combination of argument types when trying to construct it.
```

src/decompression.jl
```diff
@@ -701,54 +701,72 @@ end
...
+end
+
+function Base.getindex(B::JoinCompressed, i::Integer, j::Integer)
+    (; n, Br, Bc, row_color_ind, col_color_ind) = B
+    if i ≤ n
+        return Br[row_color_ind[j], i]
     else
-        Br_and_Bc = similar(result.Br_and_Bc, R)
+        return Bc[i - n, col_color_ind[j]]
     end
```
gdalle, line 743: Is it obvious that we win by doing this? Each `getindex` involves a dictionary lookup, which is much more expensive than for naive matrices

src/decompression.jl
```diff
@@ -701,54 +701,72 @@ end
...
-    fill!(Br_and_Bc, zero(R))
-    for c in axes(Br_and_Bc, 2)
-        if haskey(row_color_ind, c)  # some rows were colored with symmetric color c
-            copyto!(view(Br_and_Bc, 1:n, c), view(Br, row_color_ind[c], :))
-        end
-        if haskey(col_color_ind, c)  # some columns were colored with symmetric c
-            copyto!(view(Br_and_Bc, (n + 1):(n + m), c), view(Bc, :, col_color_ind[c]))
-        end
-    end
```
gdalle, line 732: Actually the previous storage wasn't so bad at all. If every color is used both for rows and for columns, there's no waste whatsoever. I think the performance comparison is not clear cut

amontoison: The waste is just that we duplicate memory Guillaume.
Why should we allocate `Br_and_Bc` and we have already `Br` and `Bc`.

src/decompression.jl
amontoison: We can also have two vectors of integer of size ncolors that di the mapping between symmetric colors and row / column colors.
It could be cheaper and faster.

src/decompression.jl
amontoison: It is a clear cut in terms of storage, for the performance I don't know with the current implementation of dictionary.

amontoison:
@gdalle 
For `row_color_ind` and `col_color_ind` we should use two `Vector{Int}` instead of `Dict{Int,Int}`.
`get_index` will be faster and we will use less memory:
```julia
julia> a = Dict{Int,Int}(i => i+1 for i = 1:100)
julia> Base.summarysize(a)
4464

julia> b = Vector{Int}(undef, 100)
julia> Base.summarysize(b)
840
```
We have a factor 5 in terms of storage.

gdalle:
The reason why they are dictionaries is that not every color is used for a row, or for a column. But we could probably put zeros as indices for unused colors

amontoison:
@gdalle I rebased the branch but I will need your help to fix issue with `compress` for bicoloring.
It could return matrices `Br` and `Bc` of different types if one of them is empty.

src/decompression.jl
```diff
@@ -701,56 +701,70 @@ end
...
+    A_and_Aᵀ = decompress(Br_and_Bc, symmetric_result)
     copyto!(A, A_and_Aᵀ[(n + 1):(n + m), 1:n])  # original matrix in bottom left corner
     return A
 end
 
 function decompress!(
     A::SparseMatrixCSC, Br::AbstractMatrix, Bc::AbstractMatrix, result::BicoloringResult
 )
-    (; large_colptr, large_rowval, symmetric_result) = result
+    (; row_color_ind, col_color_ind, symmetric_result, large_colptr, large_rowval) = result
```
amontoison, line 765:
```suggestion
    (; symmetric_to_row, symmetric_to_column, symmetric_result, large_colptr, large_rowval) = result
```

src/decompression.jl
```diff
@@ -701,56 +701,70 @@ end
...
     m, n = size(A)
-    Br_and_Bc = _join_compressed!(result, Br, Bc)
+    Br_and_Bc = JoinCompressed(m, n, Br, Bc, row_color_ind, col_color_ind)
```
amontoison, line 767:
```suggestion
    Br_and_Bc = JoinCompressed(m, n, Br, Bc, symmetric_to_row, symmetric_to_column)
```

src/decompression.jl
amontoison: fixed by #197

gdalle:
Hot damn!

amontoison:
Capture d’écran du 2025-03-04 17-02-34

I don't understand these results. I am wondering if `Br` and `Bc` are dense matrices or not in the benchmarks.
All rother results are excellent.

gdalle:
I just profiled the problematic case and we're spending most of the time in the new `getindex`, as I feared.

Screenshot 2025-03-05 at 08 53 50

gdalle:
On the other hand, in the previous version of the code (the one on `main`), I see that `Br` or `Bc` can be sparse (!!!). This means the benchmarking will be biased because they should be dense in the first place. I'll try to fix that first.

gdalle:
We also should use random ordering in the benchmarks, otherwise the results will be trivial.

gdalle:
I just merged the modifications from #198, the upcoming benchmarks will be more realistic

gdalle:
Another issue with the benchmarks is that, in the current state, we're unable to pre-allocate `Br_and_Bc` due to mismatched `decompression_eltype`. #199 fixes that

src/decompression.jl
```diff
@@ -717,56 +717,72 @@ end
...
+       AbstractMatrix{R}
+    m::Int
+    n::Int
+    Br::M1
+    Bc::M2
+    symmetric_to_row::Vector{Int}
+    symmetric_to_column::Vector{Int}
+end
+
+function Base.getindex(B::JoinCompressed, i::Int, j::Int)
```
amontoison, line 752:
```suggestion
function Base.getindex(B::JoinCompressed{R}, i::Int, j::Int) where R
```

src/decompression.jl
```diff
@@ -717,56 +717,72 @@ end
...
-            copyto!(
-                view(Br_and_Bc, (n + 1):(n + m), c), view(Bc, :, symmetric_to_column[c])
-            )
-        end
+        return Bc[i - n, symmetric_to_column[j]]
     end
-    return Br_and_Bc
+end
+
+function Base.getindex(B::JoinCompressed, k::Int)
```
amontoison, line 761:
```suggestion
function Base.getindex(B::JoinCompressed{R}, k::Int) where R
```

amontoison:
@gdalle 
Capture d’écran du 2025-03-17 02-42-13
Capture d’écran du 2025-03-17 02-42-35
Note that I still need to do an efficient decompression for acyclic bicoloring.

amontoison:
@gdalle The PR is finally ready for review!

PR #197, opened 2025-02-22
Optimize remap_colors

### LINKED ISSUE #172: Add an inplace version for remap_colors
At the end of the postprocessing, we could use an inplace version of `remap_colors` to overwrite `colors` and not allocate `remapped_colors`.

We could also avoid some allocations with the use of `view` in the following lines:
https://github.com/gdalle/SparseMatrixColorings.jl/blob/main/src/result.jl#L538-L539

amontoison:
close #172 
- Avoid the call to remap_colors in postprocess!
- Optimize `remap_colors`
- Use two vectors of integer `symmetric_to_column` and `symmetric_to_row` for the mapping between symmetric colors and row / column colors

### DIFF
--- a/src/coloring.jl
+++ b/src/coloring.jl
@@ -123,7 +123,8 @@ function star_coloring(g::AdjacencyGraph, order::AbstractOrder; postprocessing::
     end
     star_set = StarSet(star, hub, nb_spokes)
     if postprocessing
-        postprocess!(color, star_set, g)
+        # Reuse the vector forbidden_colors to compute offsets during post-processing
+        postprocess!(color, star_set, g, forbidden_colors)
     end
     return color, star_set
 end
@@ -351,7 +352,8 @@ function acyclic_coloring(g::AdjacencyGraph, order::AbstractOrder; postprocessin
     end
     tree_set = TreeSet(forest, nb_vertices(g))
     if postprocessing
-        postprocess!(color, tree_set, g)
+        # Reuse the vector forbidden_colors to compute offsets during post-processing
+        postprocess!(color, tree_set, g, forbidden_colors)
     end
     return color, tree_set
 end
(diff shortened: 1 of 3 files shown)


src/coloring.jl
```diff
@@ -642,15 +645,29 @@ function postprocess!(
...
+                num_colors_useless += 1
+                offsets[i] = -num_colors_useless
+            end
+        end
+
+        # assign the neutral color to every vertex with a useless color and remap the colors
         for i in eachindex(color)
             ci = color[i]
-            if !color_used[ci]
+            if offsets[ci] < 0
```
amontoison, line 663:
```suggestion
            !color_used[ci]
```

src/coloring.jl
```diff
@@ -642,15 +645,29 @@ function postprocess!(
...
-        # assign the neutral color to every vertex with a useless color
+        num_colors_useless = 0
+
+        # determine what are the useless colors and compute the offsets
+        for i in 1:nb_colors
+            if color_used[i]
+                offsets[i] = num_colors_useless
+            else
+                num_colors_useless += 1
+                offsets[i] = -num_colors_useless
```
amontoison, line 656:
```suggestion
```

amontoison: We don't need to compute the `offset` for neutral colors because we don't use it.

src/coloring.jl
```diff
@@ -642,15 +645,28 @@ function postprocess!(
...
+            else
+                num_colors_useless += 1
+            end
+        end
+
+        # assign the neutral color to every vertex with a useless color and remap the colors
         for i in eachindex(color)
             ci = color[i]
-            if !color_used[ci]
+            !color_used[ci]
```
amontoison, line 662:
```suggestion
            if !color_used[ci]
```

gdalle (approved):
looks good!

PR #198, opened 2025-03-05
Type-stable compress and random order in benchmarks

gdalle:


src/decompression.jl
```diff
@@ -74,14 +82,22 @@ function compress(
     row_group = row_groups(result)
     column_group = column_groups(result)
     if isempty(row_group)
-        Br = similar(A, 0, size(A, 2))
+        # ensure we get a Matrix and not a SparseMatrixCSC
+        Br_model = stack([Int[]]; dims=1) do g
+            dropdims(sum(A[g, :]; dims=1); dims=1)
+        end
+        Br = similar(Br_model, 0, size(A, 2))
```
amontoison, line 89: `Br` and `Bc` can have a different `eltype` with this modification.
If one partition is empty (row / column) and the matrix is of type Matrix{Float64}, we can have Br or Bc of type Matrix{Int}.

Am i wrong?

gdalle: I think you're wrong, the eltype is dictated by whatever is inside of `A`, not by the `[Int[]]` we provide to tell which columns to sum.
Note that this also solves our problem with `Bool` vs `Int` eltypes

gdalle:
```julia
julia> using SparseArrays

julia> A = sprand(10, 10, 0.1)
10×10 SparseMatrixCSC{Float64, Int64} with 13 stored entries:
  ⋅         ⋅    ⋅         ⋅         ⋅   0.756882  0.830887   ⋅          ⋅    ⋅ 
  ⋅         ⋅    ⋅         ⋅         ⋅    ⋅         ⋅        0.366633    ⋅    ⋅ 
  ⋅         ⋅   0.761644   ⋅         ⋅   0.376413   ⋅         ⋅          ⋅    ⋅ 
  ⋅         ⋅    ⋅        0.261975   ⋅    ⋅        0.55414   0.693315    ⋅    ⋅ 
  ⋅         ⋅   0.947163   ⋅         ⋅    ⋅         ⋅         ⋅          ⋅    ⋅ 
  ⋅         ⋅    ⋅        0.424692   ⋅    ⋅         ⋅         ⋅          ⋅    ⋅ 
  ⋅         ⋅    ⋅         ⋅         ⋅    ⋅         ⋅         ⋅          ⋅    ⋅ 
  ⋅         ⋅    ⋅         ⋅         ⋅    ⋅        0.208263   ⋅          ⋅    ⋅ 
 0.512624   ⋅    ⋅         ⋅         ⋅    ⋅         ⋅         ⋅          ⋅    ⋅ 
  ⋅         ⋅    ⋅         ⋅         ⋅    ⋅         ⋅        0.0308632   ⋅    ⋅ 

julia> B_model = stack([Int[]]; dims=1) do g
           dropdims(sum(A[g, :]; dims=1); dims=1)
       end
1×10 Matrix{Float64}:
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
```

amontoison: You're right. I mixed up things. All good for me.

Issue #67, opened 2024-08-12
Decompression from vector of colors?


gdalle:
Will never be used in practice with `GreedyColoringAlgorithm`, but one could imagine optimal colorings with JuMP, which would only output a vector of colors.

#71 contains a `LinearSystemColoringResult`, which is part of the answer in the symmetric case.

gdalle:
For star and acyclic colorings, we can actually reverse-engineer the set of stars / streets because they are 2-colored!!

amontoison:
Is it not possible that a color is used in different stars?
For the acyclic coloring, a color can be used in different trees.

gdalle:
Yes, but if we iterate over all pairs of colors we get each tree or star exactly once. Quoting from the decompression paper:

> A star coloring of a graph is a distance-1 coloring where, in addition, every path in the graph on four vertices (P4) is required to use at least three colors. An acyclic coloring of a graph is a distance-1 coloring with the further restriction that every cycle in the graph uses at least three colors. The names star and acyclic coloring are due to the structure of twocolored induced subgraphs. In a star-colored graph, a subgraph induced by any two color classes—sets of vertices having the same color—is a collection of stars. A star is a complete bipartite graph in which one of the vertex sets consists of a single vertex, called the hub. The other vertices of the star are spokes. An edge is a degenerate case of a star, in which one vertex is the hub and the other, the spoke, assigned arbitrarily. Similarly, in an acyclically colored graph, a subgraph induced by any two color classes is a collection of trees, and thus acyclic.

PR #202, opened 2025-03-08
Chordal coloring algorithm.

samuelsonric:
Chordal graphs can be minimally colored in linear time, and every such coloring is acyclic. [This paper](https://pubsonline.informs.org/doi/abs/10.1287/ijoc.1080.0286) discusses chordal graphs in §4 and suggests that they arise in practice. This patch adds an algorithm `ChordalColoringAlgorithm` that handles this case. Here are benchmarks for acyclic coloring of a band graph (defined in the paper).

```julia
using BandedMatrices, BenchmarkTools, Random, SparseArrays, SparseMatrixColorings

n = 10000 # vertices
m = 400   # bandwidth
matrix = sparse(brand(n, n, div(m, 2), 0))
matrix .+= matrix'
perm = randperm(n)
permute!(matrix, perm, perm)
problem = ColoringProblem(; structure=:symmetric, partition=:column)
```

**greedy coloring**

```julia-repl
julia> algorithm = GreedyColoringAlgorithm(NaturalOrder())
GreedyColoringAlgorithm{:direct, NaturalOrder}(NaturalOrder(), false)

julia> ncolors(coloring(matrix, problem, algorithm))
487

julia> @benchmark coloring(matrix, problem, algorithm)
BenchmarkTools.Trial: 1 sample with 1 evaluation per sample.
 Single result which took 21.141 s (0.47% GC) to evaluate,
 with a memory estimate of 329.99 MiB, over 3668722 allocations.
```

**chordal coloring**

```julia-repl
julia> algorithm = ChordalColoringAlgorithm()
ChordalColoringAlgorithm()

julia> ncolors(coloring(matrix, problem, algorithm))
201

julia> @benchmark coloring(matrix, problem, algorithm)
BenchmarkTools.Trial: 122 samples with 1 evaluation per sample.
 Range (min … max):  39.282 ms … 44.836 ms  ┊ GC (min … max): 0.00% … 1.65%
 Time  (median):     40.740 ms              ┊ GC (median):    1.13%
 Time  (mean ± σ):   40.990 ms ±  1.149 ms  ┊ GC (mean ± σ):  1.13% ± 0.83%

         ▂▂▂▆▂▂▂ ▂▆▂ █            ▂        ▄                   
  ██▆██▄▆█████████████▆▄█▆▆▄▆█▆▁▄▁█▆▄▄▆▁▆▆██▄▄▁▄▁▁▁▁▄▁▄▁▁▁▁▁▄ ▄
  39.3 ms         Histogram: frequency by time        44.2 ms <

 Memory estimate: 16.11 MiB, allocs estimate: 49.
```

The algorithm errors if the graph is not chordal.

```julia-repl
julia> invp = invperm(perm);

julia> matrix[invp[1], invp[3]] = matrix[invp[3], invp[1]] = matrix[invp[4], invp[2]] = matrix[invp[2], invp[4]] = 0;

julia> dropzeros!(matrix);

julia> ncolors(coloring(matrix, problem, ChordalColoringAlgorithm()))
ERROR: Matrix does not have a chordal sparsity pattern.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] coloring(A::SparseMatrixCSC{Float64, Int64}, ::ColoringProblem{:symmetric, :column}, algo::ChordalColoringAlgorithm)
   @ SparseMatrixColorings ~/Source/git/SparseMatrixColorings.jl/src/chordal.jl:20
 [3] top-level scope
   @ REPL[18]:1
```

If you like the algorithm then I can add tests and so forth.


samuelsonric:
[Related?](https://github.com/gdalle/SparseMatrixColorings.jl/issues/65)

samuelsonric:
Oops! I'll rerun & update the benchmarks.

gdalle:
Hi @samuelsonric, thank you for this suggestion!

Indeed, I had skipped the experimental section when reading _Efficient Computation of Sparse Hessians Using Coloring and Automatic Differentiation_ (Gebremehdin et al., 2008), but such special cases deserve our attention. Banded matrices in particular are very relevant for SciML applications, and they generate are a special case of chordal graphs.

Here are a few remarks:
- Adding a new dependency to SparseMatrixColorings.jl is rather a big deal, since it affects many many users down the line. The only way we would be willing to consider it is in a package extension.
- New coloring algorithms which are not based on two-colored structures will not be compatible with our current decompression utilities. To test your code, we would need to solve #67 first.
- From a user perspective, it might be rather confusing to see the number of coloring algorithms grow to include one for several families of graphs / matrices. What I had in mind to solve #65 was something like an `OptimalColoringAlgorithm`, which would work based on dispatch to associate matrices with known optimal colorings (like `BandedMatrix` from [BandedMatrices.jl](https://github.com/JuliaLinearAlgebra/BandedMatrices.jl)).
- Another alternative would be to encode the chordal aspect in the vertex ordering subroutine, instead of the global coloring algorithm. From what I understand, optimal coloring of a chordal graph can be achieved with a perfect elimination order. Is that what you're computing with `CliqueTrees.permutation`?

samuelsonric:
Yes: `CliqueTrees.permutation` computes a peo!

gdalle:
> Yes: CliqueTrees.permutation computes a peo!

I had the vague notion that this was an NP-hard problem?

samuelsonric:
Ok: I think a package extension makes sense. A chordal routine could easily be integrated into an `OptimalColoringAlgorithm`. Although chordality cannot be verified at compile time, it can be verified almost instantaneously at runtime using `CliqueTrees.ischordal` or `CliqueTrees.isperfect`.

samuelsonric:
> I had the vague notion that this was an NP-hard problem?

This is a bit subtle. `CliqueTrees.permutation` computes a vertex ordering in linear time. If the graph is chordal, then the ordering is a PEO. The whether the ordering is in fact perfect can also be determined in linear time.

This is the standard strategy for chordality testing. Computing a PEO for general graphs is NP hard!

gdalle:
My suggestion would be to
- define and document a new `struct PerfectEliminationOrder end` in `src/order.jl`
- implement `SparseMatrixColoring.vertices(g, ::PerfectEliminationOrder)` in `ext/SparseMatrixColoringsCliqueTrees.jl`
- add tests in `test/chordal.jl`

This seems like a minimum-effort way to add this functionality without fully solving #67, because it would be inserted within the existing `GreedyColoringAlgorithm`. What do you think?

samuelsonric:
Added a `PerfectEliminationOrder` algorithm. However, the number of colors generated by the greedy ordering algorithm seems to consistently be twice the chromatic number. Is this expected?

```jula
using BandedMatrices, BenchmarkTools, Random, SparseArrays, SparseMatrixColorings
n = 50 # vertices
m = 10 # bandwidth
matrix = sparse(brand(n, n, div(m, 2), 0))
matrix .+= matrix'
perm = randperm(n)
permute!(matrix, perm, perm)
```

```julia-repl
julia> algorithm = GreedyColoringAlgorithm(LargestFirst());

julia> ncolors(coloring(matrix, problem, algorithm))
13
```

```julia-repl
julia> import CliqueTrees

julia> algorithm = GreedyColoringAlgorithm(PerfectEliminationOrder());

julia> ncolors(coloring(matrix, problem, algorithm))
11
```

```julia-repl
julia> using SimpleGraphs, SimpleGraphAlgorithms

julia> chromatic_number(UndirectedGraph(matrix))
6
```

samuelsonric:
Maybe I selected the wrong problem type?

samuelsonric:
Added tests!

ext/SparseMatrixColoringsCliqueTreesExt.jl
```diff
@@ -0,0 +1,36 @@
...
+
+using CliqueTrees: CliqueTrees
+using SparseArrays
+using SparseMatrixColorings:
+    SparseMatrixColorings, AdjacencyGraph, BipartiteGraph, PerfectEliminationOrder, pattern
+
+function SparseMatrixColorings.vertices(
+    g::AdjacencyGraph{T}, order::PerfectEliminationOrder
+) where {T}
+    S = pattern(g)
```
gdalle, line 11: This matrix `S = pattern(g)` may have nonzeros on the diagonal, but those nonzeros are ignored when building the adjacency graph `g` (no self-loops)

test/order.jl
```diff
@@ -110,3 +113,18 @@ end;
         end
     end
 end;
+
+@testset "PerfectEliminationOrder" begin
+    problem1 = ColoringProblem(; structure=:symmetric, partition=:column)
+    problem2 = ColoringProblem(; structure=:nonsymmetric, partition=:column)
```
gdalle, line 119: Nonsymmetric problems don't work with decompression by substitution.

ext/SparseMatrixColoringsCliqueTreesExt.jl
```diff
@@ -0,0 +1,36 @@
...
+    return reverse!(order)
+end
+
+function SparseMatrixColorings.vertices(
+    bg::BipartiteGraph{T}, ::Val{side}, order::PerfectEliminationOrder
+) where {T,side}
+    S = pattern(bg, Val(side))
+
+    # construct matrix with sparsity pattern S
+    M = SparseMatrixCSC{Bool,T}(size(S)..., S.colptr, rowvals(S), ones(Bool, nnz(S)))
```
gdalle, line 28: How do you encode the bipartite aspect here?

ext/SparseMatrixColoringsCliqueTreesExt.jl
```diff
@@ -0,0 +1,36 @@
+module SparseMatrixColoringsCliqueTreesExt
+
+using CliqueTrees: CliqueTrees
+using SparseArrays
```
gdalle, line 4: Use explicit imports

ext/SparseMatrixColoringsCliqueTreesExt.jl
```diff
@@ -0,0 +1,36 @@
+module SparseMatrixColoringsCliqueTreesExt
+
+using CliqueTrees: CliqueTrees
```
gdalle, line 3: You can import everything at the top, it makes it easier to see what we depend on from CliqueTrees

src/order.jl
```diff
@@ -301,3 +301,17 @@ Instance of [`AbstractOrder`](@ref) which sorts vertices from lowest to highest
...
 """
 const DynamicLargestFirst = DynamicDegreeBasedOrder{:forward,:low2high}
+
+"""
+    PerfectEliminationOrder
+
+Instance of [`AbstractOrder`](@ref) which sorts the vertices of a chordal graph in a perfect elimination order.
+
+!!! danger
+    This order is implemented as a package extension and requires loading CliqueTrees.jl.
```
gdalle, line 311:
```suggestion
    This order is implemented as a package extension and requires loading [CliqueTrees.jl](https://github.com/AlgebraicJulia/CliqueTrees.jl).
```

src/order.jl
```diff
@@ -301,3 +301,17 @@ Instance of [`AbstractOrder`](@ref) which sorts vertices from lowest to highest
 - [`DynamicDegreeBasedOrder`](@ref)
 """
 const DynamicLargestFirst = DynamicDegreeBasedOrder{:forward,:low2high}
+
+"""
+    PerfectEliminationOrder
+
+Instance of [`AbstractOrder`](@ref) which sorts the vertices of a chordal graph in a perfect elimination order.
```
gdalle, line 308: Specify behavior on non-chordal graphs. Should it error or just return a suboptimal order?

test/order.jl
```diff
@@ -110,3 +113,18 @@ end;
...
+    algorithm = GreedyColoringAlgorithm(
+        PerfectEliminationOrder(); decompression=:substitution
+    )
+
+    for (n, m) in ((800, 80), (400, 40), (200, 20), (100, 10))
+        perm = randperm(n)
+        matrix = permute!(sparse(Symmetric(brand(n, n, m, 0), :L)), perm, perm)
```
gdalle, line 126: It would be important to test on non-symmetric matrices

gdalle (changes requested):
Thanks for this initiative! Here are a few remarks of mine ;)

ext/SparseMatrixColoringsCliqueTreesExt.jl
samuelsonric: The `CliqueTrees.permutation` function ignores self-loops!

src/order.jl
samuelsonric: Suboptimal

gdalle: Okay, then we need to add it in the docstring and test it :)

ext/SparseMatrixColoringsCliqueTreesExt.jl
samuelsonric: This is assuming that `bg` is a balanced symmetric bipartite graph (i.e. an undirected graph lol). Should I just remove this option?

test/order.jl
samuelsonric: Oh, the algorithm only works for symmetric matrices (directed graph's can't be chordal!).

ext/SparseMatrixColoringsCliqueTreesExt.jl
samuelsonric: Including `SparseArrays`?

test/order.jl
gdalle: Then we only need to define it on `AdjacencyGraph`, and test it on `:symmetric` coloring problems

samuelsonric: Perfect. I was a bit confused & just guessed.

samuelsonric:
Should be everything.

samuelsonric:
The docs should build successfully, now!

ext/SparseMatrixColoringsCliqueTreesExt.jl
```diff
@@ -0,0 +1,24 @@
...
+
+function vertices(g::AdjacencyGraph{T}, order::PerfectEliminationOrder) where {T}
+    S = pattern(g)
+
+    # construct matrix with sparsity pattern S
+    M = SparseMatrixCSC{Bool,T}(size(S)..., S.colptr, rowvals(S), ones(Bool, nnz(S)))
+
+    # construct a perfect elimination order
+    # self-loops are ignored
+    # we can also use alg=LexBFS()
```
gdalle, line 16: The algorithm could perhaps become a field of the `PerfectEliminationOrder` struct?

src/order.jl
```diff
@@ -301,3 +301,17 @@ Instance of [`AbstractOrder`](@ref) which sorts vertices from lowest to highest
 - [`DynamicDegreeBasedOrder`](@ref)
 """
 const DynamicLargestFirst = DynamicDegreeBasedOrder{:forward,:low2high}
+
+"""
+    PerfectEliminationOrder
+
+A linear-time ordering code for symmetric graphs. On [chordal graphs](https://en.wikipedia.org/wiki/Chordal_graph), the code computes a perfect elimination ordering. Otherwise, it computes a suboptimal ordering.
```
gdalle, line 308:
```suggestion
Instance of [`AbstractOrder`](@ref) which computes a perfect elimination ordering when the underlying graph is [chordal](https://en.wikipedia.org/wiki/Chordal_graph). For generic graphs, it computes a suboptimal ordering.

!!! warning
    This order can only be applied for symmetric or bidirectional coloring problems, not unidirectional ones.
```

gdalle (changes requested):
Almost there!

src/order.jl
```diff
@@ -305,7 +305,10 @@ const DynamicLargestFirst = DynamicDegreeBasedOrder{:forward,:low2high}
 """
     PerfectEliminationOrder
 
-A linear-time ordering code for symmetric graphs. On [chordal graphs](https://en.wikipedia.org/wiki/Chordal_graph), the code computes a perfect elimination ordering. Otherwise, it computes a suboptimal ordering.
+Instance of [`AbstractOrder`](@ref) which computes a perfect elimination ordering when the underlying graph is [chordal](https://en.wikipedia.org/wiki/Chordal_graph). For generic graphs, it computes a suboptimal ordering.
+
+!!! warning
+    This order can only be applied for symmetric or bidirectional coloring problems, not unidirectional ones.
```
gdalle, line 311:
```suggestion
    This order can only be applied for symmetric or bidirectional coloring problems, not unidirectional ones. Furthermore, its theoretical guarantees only hold for decompression by substitution.
```

src/order.jl
```diff
@@ -301,3 +301,20 @@ Instance of [`AbstractOrder`](@ref) which sorts vertices from lowest to highest
...
 """
 const DynamicLargestFirst = DynamicDegreeBasedOrder{:forward,:low2high}
+
+"""
+    PerfectEliminationOrder
+
+Instance of [`AbstractOrder`](@ref) which computes a perfect elimination ordering when the underlying graph is [chordal](https://en.wikipedia.org/wiki/Chordal_graph). For generic graphs, it computes a suboptimal ordering.
+
+!!! warning
+    This order can only be applied for symmetric or bidirectional coloring problems, not unidirectional ones. Furthermore, its theoretical guarantees only hold for decompression by substitution.
```
gdalle, line 311:
```suggestion
    This order can only be applied for symmetric or bidirectional coloring problems. Furthermore, its optimality guarantees only hold for decompression by substitution, not for direct decompression.
```

gdalle:
Added the restriction to decompression by substitution due to Lemma 4.2 of the 2008 paper:

> If a mapping is a distance-1 coloring for a chordal graph G, then is also an acyclic coloring for G.

This means that we can compute an optimal order for acyclic coloring that way, but not for star coloring.

gdalle:
Wait, on second thought I'm not sure the order-based approach is right. It would get us an optimal distance-1 coloring if we applied a greedy distance-1 algorithm after, but we actually apply a greedy acyclic algorithm after. I struggle to see whether we have any optimality guarantees that way.

gdalle (approved):

amontoison:
Thanks @samuelsonric for this PR!
I think we should add a small example in the documentation on how to use this new ordering with a banded matrix.

PR #204, opened 2025-03-17
Try the compression of B in acyclic decompression

amontoison:


amontoison:
@gdalle We have almost no gain in terms of speed when we compress `B` in the decompression by substitution.
I close the PR.

PR #207, opened 2025-03-25
Remove some dictionaries in star and acyclic coloring

amontoison:
Quick test to see how much we gain with a dictionary.
Finding the index of an edge before this PR was `O(1)`, now it is `O(log(n))`.
It can also be `O(1) without a dictionary but it requires more modifications and I would like to know what is the gap already.

### DIFF
--- a/src/SparseMatrixColorings.jl
+++ b/src/SparseMatrixColorings.jl
@@ -39,7 +39,8 @@ using SparseArrays:
     nzrange,
     rowvals,
     sparse,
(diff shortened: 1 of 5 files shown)


gdalle:
Putting this on hold while we think of an integer-based approach to edge identification

amontoison:
Not relevant anymore since we know how to avoid the use of a sparse strict upper triangular matrix / dictionary to do the mapping between edges and their index.

PR #209, opened 2025-03-25
Improve the decompression of star coloring for AbstractMatrix

amontoison:
Use `compressed_indices` from the `StarColoringResult`.


gdalle (approved):
really good idea!

PR #211, opened 2025-03-25
Improve degree and nb_edges

amontoison:
- Specialize `degree`and `nb_edges` if we have loops or not.
- Don't check all values in `degree` if it's not needed.


gdalle (changes requested):
I think we're missing coverage because our default test suite for bicoloring only uses the random order, which never computes degrees.
We might want to test in the random suite with `LargestFirst()` too?

amontoison:
Yes, good idea!

amontoison:
Simplified and added in #214

PR #216, opened 2025-03-31
Store edge index in `AdjacencyGraph` & get rid of dicts

gdalle:
Alternative take on #214


amontoison:
I think we also need to modify the functions for ordering if we modify `neighbors`.

gdalle:
Yeah this is still very much a work in progress

gdalle:
Closing in favor of #214

PR #217, opened 2025-03-31
Update decompression_single_color! for star coloring

amontoison:
It removes the need for the vector `spokes`.

### DIFF
--- a/src/decompression.jl
+++ b/src/decompression.jl
@@ -465,7 +465,7 @@ function decompress_single_color!(
     rvS = rowvals(S)
     for j in group[c]
         for k in nzrange(S, j)
-            if lower_index ≤ compressed_indices[k] ≤ upper_index
+            if lower_index <= compressed_indices[k] <= upper_index
                 l = compressed_indices[k] - offset
                 i = rvS[k]
                 if i == j
--- a/src/decompression.jl
+++ b/src/decompression.jl
@@ -455,30 +455,30 @@ function decompress_single_color!(
     result::StarSetColoringResult,
     uplo::Symbol=:F,
 )
-    (; ag, color, group, star_set) = result
-    (; hub, spokes) = star_set
+    (; ag, compressed_indices, group) = result
     (; S) = ag
     uplo == :F && check_same_pattern(A, S)
 
-    # Recover the diagonal coefficients of A
-    if has_diagonal(ag)
-        for i in group[c]
-            if !iszero(S[i, i])
-                A[i, i] = b[i]
-            end
-        end
-    end
-
-    # Recover the off-diagonal coefficients of A
-    for s in eachindex(hub, spokes)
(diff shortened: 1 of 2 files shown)


gdalle:
Very good idea! If we really don't need it anymore, can you get rid of spokes everywhere in this PR?

amontoison:
Done @gdalle

src/decompression.jl
```diff
@@ -455,30 +455,30 @@ function decompress_single_color!(
...
-                end
-                if in_triangle(j, i, uplo)
-                    A[j, i] = b[i]
+    offset = (c - 1) * S.n
+    lower_index = offset + 1
+    upper_index = c * S.n
+    rvS = rowvals(S)
+    for j in group[c]
+        for k in nzrange(S, j)
+            if lower_index ≤ compressed_indices[k] ≤ upper_index
```
gdalle, line 468:
```suggestion
            if lower_index <= compressed_indices[k] <= upper_index
```
Following BlueStyle

src/decompression.jl
```diff
@@ -455,30 +455,30 @@ function decompress_single_color!(
...
-                end
-                if in_triangle(j, i, uplo)
-                    A[j, i] = b[i]
+    offset = (c - 1) * S.n
+    lower_index = offset + 1
+    upper_index = c * S.n
+    rvS = rowvals(S)
+    for j in group[c]
+        for k in nzrange(S, j)
+            if lower_index <= compressed_indices[k] <= upper_index
```
gdalle, line 468: I'm not sure why this `if` is necessary, isn't the condition always true?

src/decompression.jl
```diff
@@ -455,30 +455,30 @@ function decompress_single_color!(
...
+                else
+                    # Recover the off-diagonal coefficients of A
+                    if in_triangle(i, j, uplo)
+                        A[i, j] = b[l]
+                    end
+                    if in_triangle(j, i, uplo)
+                        A[j, i] = b[l]
+                    end
                 end
             end
```
gdalle, line 483:
```suggestion
            l = compressed_indices[k] - offset
            i = rvS[k]
            if i == j
                # Recover the diagonal coefficients of A
                A[i, i] = b[l]
            else
                # Recover the off-diagonal coefficients of A
                if in_triangle(i, j, uplo)
                    A[i, j] = b[l]
                end
                if in_triangle(j, i, uplo)
                    A[j, i] = b[l]
                end
            end
```

src/decompression.jl
amontoison: No, it is not always true.
What we do is that we iterate over all `A[i,j]` and `A[j,i]` and we could use the color of `i` to recover the coefficient.

gdalle: My bad

gdalle (approved):

PR #218, opened 2025-03-31
Add a default value for postprocessing

amontoison:
I checked something a few days with `@code_warntype` and I found that the compiler generated a very complex code to handle the lack of default value for the keyword argument.

It could also be an argument.

### DIFF
--- a/src/coloring.jl
+++ b/src/coloring.jl
@@ -294,7 +294,9 @@ If `postprocessing=true`, some colors might be replaced with `0` (the "neutral" 
 
 > [_New Acyclic and Star Coloring Algorithms with Application to Computing Hessians_](https://epubs.siam.org/doi/abs/10.1137/050639879), Gebremedhin et al. (2007), Algorithm 3.1
 """
-function acyclic_coloring(g::AdjacencyGraph, order::AbstractOrder; postprocessing::Bool=false)
+function acyclic_coloring(
+    g::AdjacencyGraph, order::AbstractOrder; postprocessing::Bool=false
+)
     # Initialize data structures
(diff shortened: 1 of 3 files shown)


gdalle:
Actually, not putting a default value in the inner method was a conscious choice. It helps prevent situations where you forget to pass the keyword argument from the outer method, thus ending up with a default value you didn't choose.
We can make it a positional argument but honestly this is so far down the list of suboptimalities that I don't think it matters (otherwise JET would have yelled at us)

gdalle:
And there's also an upside to having a setting always be a keyword or positional argument throughout the code, instead of switching. Keyword arguments are there to simplify writing, so it's okay to use them

gdalle (commented):
I think this doesn't change anything and I'd rather preserve the current behavior

gdalle:
Yes it is okay, my objection was two-fold:
1) it changes the logic of inner calls compared to outer calls
2) it brings no measurable performance benefit
Of course both are rather minor objections in this specific case, which is a very simple PR, but I'm mentioning them anyway.

If we don't try to maintain a simple and readable library, we will end up like ColPack.
I'm all for optimizing things which we know have a performance impact. But that requires profiling bottlenecks first, and working on those.

gdalle:
Essentially I'm protesting the spirit of the PR, more than the specific code inside it ^^

amontoison:
I just removed the need to specify `postprocessing`, which was mandatory because no default value where available.
I think it is good for everyone now (user and compiler).

gdalle (approved):
Again, I'm not saying this is a bad idea, I'm saying this is a very you idea ;)

PR #222, opened 2025-04-01
Propagate arbitrary integer types as indices

gdalle:
The idea is that if we propagate integer types carefully, we could store all indices in `Int32` whenever that is enough, leading to a 2x reduction of memory footprint.

Caveats:

- Just because an integer type `T` is able to store the vector of colors doesn't mean it is able to store every vertex.
- As a result, I use separate integer types for indices (often called `T1`) and colors (often called `T2`)

Still a draft, don't merge yet because I fear this might make some existing code error. Also needs tests on `Int32` indices.

### DIFF
--- a/src/coloring.jl
+++ b/src/coloring.jl
@@ -87,7 +87,7 @@ function star_coloring(
     first_neighbor = fill((zero(T), zero(T), zero(T)), nv)  # at first no neighbors have been encountered
     treated = zeros(T, nv)
     star = Vector{T}(undef, ne)
-    hub = Int[]  # one hub for each star, including the trivial ones
+    hub = T[]  # one hub for each star, including the trivial ones
     vertices_in_order = vertices(g, order)
 
     for v in vertices_in_order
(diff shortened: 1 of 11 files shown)


src/coloring.jl
```diff
@@ -367,23 +371,23 @@ Encode a set of 2-colored trees resulting from the [`acyclic_coloring`](@ref) al
...
 """
-struct TreeSet
-    reverse_bfs_orders::Vector{Vector{Tuple{Int,Int}}}
+struct TreeSet{T}
+    reverse_bfs_orders::Vector{Vector{Tuple{T,T}}}
     is_star::Vector{Bool}
 end
 
-function TreeSet(g::AdjacencyGraph, forest::Forest{Int})
+function TreeSet(g::AdjacencyGraph{T}, forest::Forest) where {T}
```
amontoison, line 379:
```suggestion
function TreeSet(g::AdjacencyGraph{T}, forest::Forest{T}) where {T}
```

PR #224, opened 2025-04-02
feat: allow full direct decompression into larger SparseMatrixCSC

gdalle:
This is an experimental fix for https://github.com/SciML/OrdinaryDiffEq.jl/issues/2653#issuecomment-2772595854, to be discussed and tested


gdalle:
@amontoison moving this from "enabled by default" to "enabled on demand" took a bit more work than expected, and I had to add a lot more tests to be confident. I wouldn't mind another round of review

src/decompression.jl
```diff
@@ -354,12 +382,28 @@ function decompress_single_color!(
...
+        for k in eachindex(compressed_indices)
+            nzA[k] = B[compressed_indices[k]]
+        end
+    else  # nnz(A) > nnz(S)
+        rvA, rvS = rowvals(A), rowvals(S)
+        shift = 0
+        for j in axes(S, 2)
+            for k in nzrange(S, j)
+                i = rvS[k]
+                while (k + shift) < A.colptr[j] || rvA[k + shift] < i
```
amontoison, line 399:
```suggestion
                while (k + shift) < A.colptr[j+1] || rvA[k + shift] < i
```

src/decompression.jl
```diff
@@ -354,12 +382,28 @@ function decompress_single_color!(
...
+                    nzA[k + shift] = zero(eltype(A))
```
amontoison, line 400: Do we want to fill the other nonzeros with 0?
Maybe the other coefficients are updated before the call to `decompress!` and we overwrite them.
At least we should document what we do on the nonzeros entries not that are not in the sparsity pattern of `S`.
I imagine a case where we make Jacobian full column rank with a block `αI` below `J` that is never updated.

Example: https://jso.dev/QRMumps.jl/dev/api/#QRMumps.qrm_shifted_spmat

src/decompression.jl
```diff
@@ -416,12 +460,28 @@ function decompress_single_color!(
...
+        for k in eachindex(nzA, compressed_indices)
+            nzA[k] = B[compressed_indices[k]]
+        end
+    else  # nnz(A) > nnz(S)
+        rvA, rvS = rowvals(A), rowvals(S)
+        shift = 0
+        for j in axes(S, 2)
+            for k in nzrange(S, j)
+                i = rvS[k]
+                while (k + shift) < A.colptr[j] || rvA[k + shift] < i
```
amontoison, line 477:
```suggestion
                while (k + shift) < A.colptr[j+1] || rvA[k + shift] < i
```

src/decompression.jl
```diff
@@ -488,25 +548,54 @@ end
...
+            for k in eachindex(nzA, compressed_indices)
+                nzA[k] = B[compressed_indices[k]]
+            end
+        else  # nnz(A) > nnz(S)
+            rvA, rvS = rowvals(A), rowvals(S)
+            shift = 0
+            for j in axes(S, 2)
+                for k in nzrange(S, j)
+                    i = rvS[k]
+                    while (k + shift) < A.colptr[j] || rvA[k + shift] < i
```
amontoison, line 566:
```suggestion
                    while (k + shift) < A.colptr[j+1] || rvA[k + shift] < i
```

src/matrices.jl
```diff
@@ -61,25 +61,18 @@ function respectful_similar(A::Union{Symmetric,Hermitian}, ::Type{T}) where {T}
...
-- if the return is `false`, they definitely don't have the same sparsity pattern
-"""
-same_pattern(A, B) = size(A) == size(B)
-
-function same_pattern(
-    A::Union{SparseMatrixCSC,SparsityPatternCSC},
-    B::Union{SparseMatrixCSC,SparsityPatternCSC},
-)
-    return size(A) == size(B) && nnz(A) == nnz(B)
+same_pattern(A::AbstractMatrix, S; allow_denser::Bool=false) = true
```
amontoison, line 64: I think we should rename this function.
`compatible_sparsity_pattern` or `compatible_pattern` is probably a better name with the new option `allow_denser`.

src/decompression.jl
```diff
@@ -193,9 +193,21 @@ The out-of-place alternative is [`decompress`](@ref).
...
 For `:symmetric` coloring results (and for those only), an optional positional argument `uplo in (:U, :L, :F)` can be passed to specify which part of the matrix `A` should be updated: the Upper triangle, the Lower triangle, or the Full matrix.
 When `A isa SparseMatrixCSC`, using the `uplo` argument requires a target matrix which only stores the relevant triangle(s).
 
+Some coloring algorithms ([`GreedyColoringAlgorithm`](@ref) and [`ConstantColoringAlgorithm`](@ref)) have an option called `allow_denser`, which enables in-place decompression into a `SparseMatrixCSC` containing more structural nonzeros than the sparsity pattern used for coloring.
+
+!!! warning
+    Decompression into a denser `SparseMatrixCSC` is only implemented when all the conditions below are satisfied simultaneously:
+      - the partition is `:row` or `:column` (not `:bidirectional`)
+      - the decompression is `:direct` (not `:substitution`)
+      - the decompression is full (so [`decompress_single_color!`](@ref) will not work)
```
amontoison, line 207: You also did the modification for one triangle, this comment is a little bit misleading as the user could understand that only `uplo=:F` is supported.

src/decompression.jl
```diff
@@ -488,25 +548,54 @@ end
...
+        rvA, rvS = rowvals(A), rowvals(S)
         l = 0  # assume A has the same pattern as the triangle
+        shift = 0
         for j in axes(S, 2)
             for k in nzrange(S, j)
                 i = rvS[k]
                 if in_triangle(i, j, uplo)
                     l += 1
-                    nzA[l] = B[compressed_indices[k]]
+                    while (l + shift) < A.colptr[j] || rvA[l + shift] < i
```
amontoison, line 584:
```suggestion
                    while (l + shift) < A.colptr[j+1] || rvA[l + shift] < i
```

amontoison:
@gdalle 
I think it's fine to not use `allow_denser` after thinking about it.
We can do something like this by default to protect the users that provides sparse matrices with a sparsity pattern that doesn't match the one used for coloring.
```julia
function decompress!(A::SparseMatrixCSC, B::AbstractMatrix, result::ColumnColoringResult)
    (; compressed_indices, allow_denser) = result
    S = result.bg.S2
    nzA = nonzeros(A)
    if nnz(A) == nnz(S)
        for k in eachindex(compressed_indices)
            nzA[k] = B[compressed_indices[k]]
        end
    else  # nnz(A) > nnz(S)
        rvA, rvS = rowvals(A), rowvals(S)
        shift = 0
        for j in axes(S, 2)
            for k in nzrange(S, j)
                i = rvS[k]
                found = false
                while (k + shift) < A.colptr[j+1] && !found
                    if rvA[k + shift] == i
                        found = true
                        nzA[k + shift] = B[compressed_indices[k]]
                    else
                        shift += 1
                        nzA[k + shift] = zero(eltype(A))
                    end
                end
                !found && error("...")
            end
        end
        nzA[(nnz(S) + shift + 1):end] .= zero(eltype(A))
    end
    return A
end
```

src/decompression.jl
amontoison: I suggest to document that `decompress!` and `decompress_single_color!` don't modify the nonzeros except the ones expected during the decompression.

src/decompression.jl
amontoison: @gdalle Imagine that we we are on the last column of `A` and that this column only contains nonzeros on rows < i. We will have an error because we stop for the reason `(k + shift) == A.colptr[j+1]`, which is `nnzA+1` in that case.
```
nzA[k + shift] = B[compressed_indices[k]]  # out-of-bound
```

gdalle: I don't understand why your code fixes the issue? We want to wait until we're in the right column (`k + shift >= A.colptr[j]`) and at the right row (`rvA[k + shift] = i`) before making an actual change to `A`. 
And I think the situation you imagine relies on the provided sparsity pattern being _less dense_ than `S`, because it is missing the nonzero at `(i, j)`? That is a case we explicitly don't consider

src/decompression.jl
gdalle: Yeah we fill them with zeros, it is the most logical choice. `decompress!` means "make the target equal to the Jacobian"

src/decompression.jl
gdalle:
```suggestion
      - the decompression is for all colors (so [`decompress_single_color!`](@ref) will not work)
```

gdalle:
Following the discussion in https://github.com/SciML/OrdinaryDiffEq.jl/issues/2653#issuecomment-2778598746, this is put on hold for the time being. Let's not delete the branch in case we need it again.

PR #226, opened 2025-04-02
Fix type conversion for weird orders

gdalle:


src/order.jl
```diff
@@ -137,15 +137,17 @@ struct DegreeBuckets{T}
...
     for d in degrees
         deg_count[d + 1] += 1
     end
     # bucket limits
-    bucket_high = cumsum(deg_count)
-    bucket_low = vcat(zero(T), @view(bucket_high[1:(end - 1)]))
+    bucket_high = accumulate(+, deg_count)
+    bucket_low = similar(bucket_high)
+    bucket_low[1] = 0
+    bucket_low[2:end] .= bucket_high[1:(end - 1)]
```
amontoison, line 150: Use a view on the right side of `.=` to avoid some allocations.

PR #228, opened 2025-04-02
Improve the computation of TreeSet

amontoison:


src/coloring.jl
```diff
@@ -446,6 +456,10 @@ function TreeSet(g::AdjacencyGraph{T}, forest::Forest{T}) where {T}
     # meaning that one vertex is directly connected to all other vertices in the tree
     is_star = Vector{Bool}(undef, nt)
 
+    # Number of edges treated
+    num_edges_treated = zero(T)
+
+    # reverse_bfs_orders contains the reverse breadth first (BFS) traversal order for each tree in the forest
     for k in 1:nt
         tree = trees[k]
```
gdalle, line 464: This is the only place where `trees` is actually used. Right now it is one dictionary per tree, which is still rather wasteful and probably accounts for a large percentage of our allocs. Perhaps it could be replace with a single dictionary that gets updated for each tree when computing the degrees?

gdalle (commented):
I think we can use this opportunity to remove one more big alloc

src/coloring.jl
gdalle: On second thought, let's do it in a separate PR, since this one works

gdalle (approved):

src/coloring.jl
amontoison: I started to work on it and it is not a simple task.

PR #229, opened 2025-04-03
Allow exact reproduction of ColPack's dynamic orders

gdalle:
Judging by my benchmarks, this is so slow that I'm not even sure we should allow it. Switching from the naive buckets to the smart buckets means that imitating ColPack behavior has become very very expensive.


amontoison (approved):

amontoison:
@gdalle I think you remove the following warning now:
https://github.com/gdalle/SparseMatrixColorings.jl/blob/main/src/order.jl#L112-L113

PR #233, opened 2025-04-05
Remove the dictionaries in TreeSet

amontoison:
The code is not yet documented but we can have an idea of the gain.

### DIFF
--- a/src/coloring.jl
+++ b/src/coloring.jl
@@ -460,7 +460,6 @@ function TreeSet(
     nt_visited = 0
     rvS = rowvals(S)
     for j in axes(S, 2)
-
         for pos in nzrange(S, j)
             i = rvS[pos]
             if i != j
--- a/src/coloring.jl
+++ b/src/coloring.jl
@@ -456,10 +456,10 @@ function TreeSet(
     # Maintain a record of visited trees to efficiently reset found_in_tree
     visited_trees = Vector{T}(undef, nt)
 
+    # Number of trees visited for each column of S
+    nt_visited = 0
     rvS = rowvals(S)
     for j in axes(S, 2)
-        # Number of visited trees that contain vertex j
-        ntree_visited = 0
 
         for pos in nzrange(S, j)
             i = rvS[pos]
@@ -503,6 +503,7 @@ function TreeSet(
         for t in 1:nt_visited
             found_in_tree[visited_trees[t]] = false
         end
+        nt_visited = 0
     end
 
     # Compute a shifted cumulative sum of tree_edge_indices, starting from one
--- a/src/coloring.jl
+++ b/src/coloring.jl
@@ -451,8 +451,7 @@ function TreeSet(
     end
 
     # found_in_tree indicates if a given vertex is in each tree
-    found_in_tree = Vector{Bool}(undef, nt)
-    fill!(found_in_tree, false)
+    found_in_tree = fill(false, nt)
 
     # Maintain a record of visited trees to efficiently reset found_in_tree
     visited_trees = Vector{T}(undef, nt)
--- a/src/coloring.jl
+++ b/src/coloring.jl
@@ -293,7 +293,7 @@ function acyclic_coloring(
 
     buffer = forbidden_colors
     reverse_bfs_orders = first_visit_to_tree
-    tree_set = TreeSet(g, forest, buffer, reverse_bfs_orders)
+    tree_set = TreeSet(g, forest, buffer, reverse_bfs_orders, ne)
     if postprocessing
         # Reuse the vector forbidden_colors to compute offsets during post-processing
         offsets = forbidden_colors
@@ -385,11 +385,12 @@ function TreeSet(
     forest::Forest{T},
     buffer::AbstractVector{T},
     reverse_bfs_orders::Vector{Tuple{T,T}},
+    ne::Integer,
 ) where {T}
     S = pattern(g)
     edge_to_index = edge_indices(g)
     nv = nb_vertices(g)
-    (; nt, ranks) = forest
+    (; nt, ranks, parents) = forest
 
     # root_to_tree is a vector that maps a tree's root to the index of the tree
     # We can recycle the vector "ranks" because we don't need it anymore to merge trees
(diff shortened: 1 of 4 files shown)


amontoison:
I will finalize this PR next week after that the PR #237 is merged.

gdalle:
@amontoison can you fix the conflicts with main? I merged #237 like you had asked as a preliminary step

amontoison:
@gdalle Done, I rebased the PR.
I spend a lot of time to add comments in the code, such that you can understand I am doing.
The main idea is to build a kind of "CSC" format for all trees in the forest in one pass.
It allows to quickly know what are the vertices in each tree and also the neighbors of a vertex inside a tree.

In theory, all code of `TreeSet` should be in the `TreeSetResultColoring` but we want the vector `is_star` for the posprocessing and we only compute it during the reverse BFS of the trees.

src/coloring.jl
```diff
@@ -399,45 +400,95 @@ function TreeSet(
...
+        neighbor_position[k] = neighbor_position[k - 1] + 2 * tree_edge_indices[k]
+    end
+
+    # found_in_tree indicates if a given vertex is in each tree
+    found_in_tree = Vector{Bool}(undef, nt)
+
     rvS = rowvals(S)
     for j in axes(S, 2)
+        # Reset the vector found_in_tree to indicate that the vertex j is not present in any tree yet
+        fill!(found_in_tree, false)
```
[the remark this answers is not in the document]
amontoison: Excellent idea Guillaume!

src/coloring.jl
```diff
@@ -399,45 +400,95 @@ function TreeSet(
...
+
+    # nvmax is the number of vertices in the largest tree of the forest
+    # Note: the number of vertices in a tree is equal the number of edges plus one
+    nvmax = maximum(tree_edge_indices) + one(T)
+
+    # Vector containing the list of vertices, grouped by tree
+    # Note: the total number of edges in the graph is "ne", so there are "ne + nt" vertices across all trees
+    tree_vertices = Vector{T}(undef, ne + nt)
+
+    # Provide the positions of the first and last neighbors for each vertex in "tree_vertices", within the tree to which the vertex belongs
```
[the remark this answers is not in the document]
amontoison: The neighbors in `tree_neighbors` are the neighbors of each vertex from `tree_vertices`.
The vertices in `tree_vertices` are grouped by trees (and the same applies to `tree_neighbors`).
A vertex can appear multiple times in `tree_vertices`, but it will have different neighbors each time, as it belongs to different trees.

Based on that, do you suggest a better code comment?

src/coloring.jl
```diff
@@ -399,47 +400,109 @@ function TreeSet(
...
+
+                # Increment neighbor count for j in the tree (shifted by 1 to facilitate the final cumsum)
+                tree_neighbor_indices[vertex_index + 1] += 1
             end
         end
+
+        # Reset found_in_tree
+        for t in 1:nt_visited
+            found_in_tree[visited_trees[t]] = false
+        end
```
amontoison, line 505:
```suggestion
        end
        nt_visited = 0
```

src/coloring.jl
```diff
@@ -399,47 +400,109 @@ function TreeSet(
...
+
+    # found_in_tree indicates if a given vertex is in each tree
+    found_in_tree = fill(false, nt)
+
+    # Maintain a record of visited trees to efficiently reset found_in_tree
+    visited_trees = Vector{T}(undef, nt)
+
     rvS = rowvals(S)
     for j in axes(S, 2)
+        # Number of visited trees that contain vertex j
```
amontoison, line 461:
```suggestion
```

src/coloring.jl
```diff
@@ -399,47 +400,109 @@ function TreeSet(
...
+        ntree_visited = 0
```
amontoison, line 462:
```suggestion
```

src/coloring.jl
```diff
@@ -399,47 +400,109 @@ function TreeSet(
...
+        vertex_position[k] = vertex_position[k - 1] + tree_edge_indices[k] + 1
+        neighbor_position[k] = neighbor_position[k - 1] + 2 * tree_edge_indices[k]
+    end
+
+    # found_in_tree indicates if a given vertex is in each tree
+    found_in_tree = fill(false, nt)
+
+    # Maintain a record of visited trees to efficiently reset found_in_tree
+    visited_trees = Vector{T}(undef, nt)
+
```
amontoison, line 458:
```suggestion

    # Number of trees visited for each column of S
    nt_visited = 0
```

src/coloring.jl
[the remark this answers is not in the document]
gdalle: I think we need a better comment on the one above then, thanks

src/coloring.jl
```diff
@@ -399,47 +400,110 @@ function TreeSet(
...
+        # Update the number of edges for the current tree (shifted by 1 to facilitate the final cumsum)
+        tree_edge_indices[index_tree + 1] += 1
+    end
+
+    # nvmax is the number of vertices in the largest tree of the forest
+    # Note: the number of vertices in a tree is equal the number of edges plus one
+    nvmax = maximum(tree_edge_indices) + one(T)
+
+    # Vector containing the list of vertices, grouped by tree
```
gdalle, line 427:
```suggestion
    # Vector containing the list of vertices, grouped by tree (each vertex appears once for every tree it belongs to)
```

src/coloring.jl
```diff
@@ -399,47 +400,110 @@ function TreeSet(
...
+                # Increment neighbor count for j in the tree (shifted by 1 to facilitate the final cumsum)
+                tree_neighbor_indices[vertex_index + 1] += 1
             end
         end
+
+        # Reset found_in_tree
+        for t in 1:nt_visited
+            found_in_tree[visited_trees[t]] = false
+        end
+        nt_visited = 0
```
gdalle, line 506: I think this needs to move inside the loop on `j`?

gdalle (changes requested):

src/coloring.jl
gdalle: If I'm wrong and the tests do not error, that's a problem

gdalle (approved):

PR #239, opened 2025-04-05
Move all subroutines in star_coloring and acyclic_coloring

amontoison:
For the paper, I explained `star_coloring` and `acyclic_coloring` as one algorithm without the subroutines.
Based on our modifications and the additional arguments passed to the subroutines, I am wondering if it might be more relevant to keep just two functions.

### DIFF
--- a/src/coloring.jl
+++ b/src/coloring.jl
@@ -99,10 +99,20 @@ function star_coloring(
             if p == v  # Case 1
                 if treated[q] != v
                     # forbid colors of neighbors of q
-                    _treat!(treated, forbidden_colors, g, v, q, color)
+                    for x in neighbors(g, q)
+                        !has_diagonal(g) || (q == x && continue)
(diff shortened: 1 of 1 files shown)


gdalle:
I disagree, there are two purposes to the subroutines:

- avoiding code duplication (`treat!` is called several times for instance)
- increasing clarity by separating the conceptual steps of the algorithm

Our code does not have to exactly mirror the paper. When I suggested presenting just one algorithm (the main loop), I meant we could keep the subroutines as subroutines, without recursing into them. I didn't mean they should be unrolled into a giant pseudocode, and thus I don't think we should unroll in our code either. Not to mention it mirrors the structure of the original paper.

amontoison:
It's fine for me. It was just a suggestion.
I'm closing the PR.
I just changed one comment but I will open another PR for that.

PR #241, opened 2025-04-08
Speed up LargestFirst order

gdalle:
- Speed up degree computation in adjacency graphs with a diagonal
- Compute the degree of each vertex once and store it before sorting
- Avoid `fill!(visited, false)` in other places when only some coefficients have been set to `true`


gdalle:
Merging since this is a very easy one

PR #246, opened 2025-04-10
Add ordering benchmarks

gdalle:
No need to run benchmarks on this one since the test cases are defined on main, we can merge right away if it looks ok

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -51,6 +51,7 @@ for structure in [:nonsymmetric, :symmetric],
 end
 
 for structure in [:nonsymmetric, :symmetric],
+    partition in (structure == :nonsymmetric ? [:column, :row] : [:column]),
     order in [LargestFirst(), SmallestLast(), IncidenceDegree(), DynamicLargestFirst()],
     n in [10^3, 10^5],
     p in [2 / n, 5 / n, 10 / n]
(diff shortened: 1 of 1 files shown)


benchmark/benchmarks.jl
```diff
@@ -44,4 +45,34 @@ for structure in [:nonsymmetric, :symmetric],
...
+        bench_ord = @benchmarkable begin
+            for g in $gs
+                SMC.vertices(g, $order)
+            end
+        end
+    else
+        gs = [SMC.BipartiteGraph(A) for A in As]
+        bench_ord = @benchmarkable begin
+            for g in $gs
+                SMC.vertices(g, Val(1), $order)
```
amontoison, line 71: I think we should split the elapsed time for ordering of the rows and the columns?

benchmark/benchmarks.jl
```diff
@@ -44,4 +45,34 @@ for structure in [:nonsymmetric, :symmetric],
...
+                SMC.vertices(g, Val(2), $order)
+            end
+        end
+    end
+
+    SUITE[:order][structure][string(order)]["n=$n"]["p=$p"] = bench_ord
```
amontoison, line 77:
```
SUITE[:order][structure][partition][string(order)]["n=$n"]["p=$p"] = bench_ord
```

PR #248, opened 2025-04-10
Remove some useless operations in dynamic orders

gdalle:


test/order.jl
```diff
@@ -97,6 +97,8 @@ end;
         IncidenceDegree(; reproduce_colpack=true),
         DynamicLargestFirst(),
         DynamicLargestFirst(; reproduce_colpack=true),
+        DynamicDegreeBasedOrder{:forward,:high2low}(),
```
amontoison, line 100: Is it equivalent to `LF`?

amontoison: The answer is no.

amontoison (approved):

PR #250, opened 2025-04-24
Use inbounds in partial_distance2_coloring

amontoison:
I got a big difference on my ANL laptop, I don't know if it is related to the CPU or the `@inbounds`.
I will check with our benchmarks.

### DIFF
--- a/src/coloring.jl
+++ b/src/coloring.jl
@@ -38,14 +38,15 @@ function partial_distance2_coloring!(
     for v in vertices_in_order
         for w in neighbors(bg, Val(side), v)
             for x in neighbors(bg, Val(other_side), w)
-                if !iszero(color[x])
-                    forbidden_colors[color[x]] = v
+                @inbounds c = color[x]
+                if !iszero(c)
+                    @inbounds forbidden_colors[c] = v
                 end
             end
         end
         for i in eachindex(forbidden_colors)
             if forbidden_colors[i] != v
-                color[v] = i
+                @inbounds color[v] = i
                 break
             end
         end


amontoison:
@gdalle On the left it is with the modifications of this PR.
On the right it is the version on `main`.
Capture d’écran du 2025-04-23 23-31-00

gdalle:
My stance on `@inbounds` hasn't changed: the unsafety is not worth the speed bump. I spent 2h yesterday fighting with an Enzyme segfault, which can't happen with our packages precisely because we don't use that macro. So let's keep it that way.

Even if we actually wanted to use it, we would need tests beforehand to guarantee that the indices are indeed within the bounds (that's what base Julia does for dense linalg). For this sparse case, it would require iterating through all the row values to verify that they are in the correct interval, so the check has complexity `O(nnz)`.

Besides, speeding up coloring was useful for the paper to compare with ColPack, but in the real world, this kind of coloring is fast enough that no one will bat an eye anyway.

I don't have many red lines, but unsafe memory accesses and aliasing is the biggest one.

amontoison:
Enzyme is another beast and I totally agree with you that we need to put some limit sometimes.

I agree that people can still do crazy things and break the sparsity pattern of a CSC matrix (but a very small %).
Maybe we should add a comment to say that the difference is only related to bound check but my issue is that some loops are not vectorized because of that.
I think it is still valuable to explain the gap with C++.

Never do programming in low-level languages Guillaume, we don't have bound check in C or Fortran.
You need ASAN, MSAN, UBSAN (or Valgrind) to track errors. Always a pleasure. :)

gdalle:
The fact that SMC is coded in a high-level language and memory-safe by default is an asset, not a curse. I'll happily add a couple lines to the paper explaining that, but I don't want to compromise the safety of our software to get these last percentage points and match ColPack. If someone inadvertently puts a -1 in the rowvals while constructing their sparse matrix, they should get an error, not a REPL crash with random corruptions of memory.

gdalle:
That being said, the present experiment is valuable and I thank you for running it! I just think it belongs in the paper as a comment, like you suggested, but not in the production-ready software. SciML and friends will already find a thousand ways to misuse our code, let's not give them one that is actually dangerous.

amontoison:
I agree with you 🙂
I was curious to understand what could bring the difference of performance in the "simple" partial distance-2 coloring.
It is just an optimization that can't the compile because of the bound checks.

amontoison:
Another alternative is to add a macro `@inbounds_smc`   that is a no-op and nobody can modify it except us.
It is hardcoded in SMC.jl but we can change it behavior for some benchmarks.

gdalle:
Let's stick with the comment then

amontoison:
I close the PR.

Conclusion: The compiler can't vectorize some operations because of bound checks. We want to keep them in SMC.jl to ensure that we can easily detect a mistake from the user.

We will add a comment about that in the paper (C++ vs Julia).

gdalle:
Note that we haven't proven it's due to lack of vectorization. It may also be that the bounds checking itself takes significant time. This could be detected by profiling.

PR #255, opened 2025-06-23
Support sparse matrices from rocSPARSE

### LINKED ISSUE #254: Support rocSPARSE matrices
It should be very similar to #253.

amontoison:
close #254 
Very similar to #253.


gdalle:
This seems to be just in the `compress` function which is not performance-critical. Maybe as a first step you can do `SparseMatrixCSC(Matrix(A_gpu))`?

PR #263, opened 2025-10-08
Make any coloring algorithm compatible with SMC

### LINKED ISSUE #67: Decompression from vector of colors?

gdalle:
Fixes #67

**Context**

At the moment, DI can only use algorithms from SMC, but people might want to code other coloring algorithms following the [ADTypes interface](https://docs.sciml.ai/ADTypes/stable/#Coloring-algorithm), for instance our friend @SouthEndMusic who was interested in provably optimal colorings. During decompression, we will still need SMC result types, but those are deeply tied to the coloring algorithm itself (for star and acyclic colorings + their bidirectional counterparts). How do we create SMC results from a generic algorithm which outputs only a vector of colors?

**Idea**

Pretend we run the greedy algorithm, but force the color choices and only piggyback the rest of the procedure for constructing data structures like the `StarSet` and `TreeSet`. This enables maximum code reuse.

**Todo**

- [ ] Test on a non-trivial coloring algorithm (not just `ADTypes.NoColoringAlgorithm`)
- [ ] Solve https://github.com/SciML/ADTypes.jl/issues/69 and then integrate that
- [ ] Figure out whether we can accept zero (unused) colors in the output of `ADTypes.symmetric_coloring`

### DIFF
--- a/src/adtypes.jl
+++ b/src/adtypes.jl
@@ -1,21 +1,32 @@
-function coloring(
-    A::AbstractMatrix,
-    ::ColoringProblem{:nonsymmetric,:column},
-    algo::ADTypes.NoColoringAlgorithm;
-    kwargs...,
-)
-    bg = BipartiteGraph(A)
-    color = convert(Vector{eltype(bg)}, ADTypes.column_coloring(A, algo))
-    return ColumnColoringResult(A, bg, color)
-end
+## From ADTypes to SMC
 
 function coloring(
     A::AbstractMatrix,
-    ::ColoringProblem{:nonsymmetric,:row},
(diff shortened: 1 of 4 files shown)


gdalle:
Gentle bump @amontoison, this one is ready for you while #267 marinates.

PS: I deactivated the formatting test because it passes locally and I can't figure out what's wrong. More generally, JuliaFormatter seems less and less reliable, so I'll soon switch to Runic anyway.

PR #267, opened 2025-10-14
Revisit the post-processing 

### LINKED ISSUE #264: Bicoloring postprocessing seems to favor removing column colors and not row colors
... which is bad for autodiff since reverse passes are more expensive than forward passes.

I wonder if this could be due to the order in which we iterate through the stars / trees? It doesn't seem to be influenced by the order in which we greedily color vertices.

```julia
A = diagm(ones(Bool, 3))
problem = ColoringProblem{:nonsymmetric,:bidirectional}()
algo = GreedyColoringAlgorithm(RandomOrder(); postprocessing=true)
for _ in 1:10
    result = coloring(A, problem, algo)
    @show column_colors(result), row_colors(result)
end
```

Perhaps in this line

https://github.com/gdalle/SparseMatrixColorings.jl/blob/7efc6bc7650c442c326aa40fe3063d27f6bd8ea5/src/coloring.jl#L646

we could use the computed vertex order instead, or its reverse order? Not sure which one makes more sense?

I noticed this because in https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/864 I sometimes got empty forward groups but never empty reverse groups.

@amontoison

amontoison:
close #264


src/coloring.jl
```diff
@@ -653,12 +655,47 @@ function postprocess!(
...
+                for k in nzrange(S, j)
+                    i = rvS[k]
+                    if i > j
+                        index_ij = edge_to_index[k]
+                        s = star[index_ij]
+                        h = hub[s]
+                        # The hub of this trivial star is still unknown
+                        if h < 0
+                            if neutralized_first == :rows
+                                # j represents a column in the context of bicoloring
```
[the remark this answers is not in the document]
amontoison: LGTM, but how do we specify that is a unidirectional case? Shoud we add a boolean argument?

gdalle:
Or we pass an additional argument to the coloring

amontoison:
> We could store this information in the adjacency graph, to know whether it comes from a symmetric matrix or not, and in the latter case have access to the row/column split

I prefer this approach because we can know that we have "augmented adjacency matrix" this way, and thus know that we are in bicoloring context for the function `postprocess!`.

I will not need to pass the boolean parameter `bicoloring` .

gdalle:
I suggest replacing `AdjacencyGraph` with:

```julia
struct AdjacencyGraph{T<:Integer,bicoloring}
    S::SparsityPatternCSC{T}
    edge_to_index::Vector{T}
    original_size::Tuple{Int,Int}
end
```

We no longer need `has_diagonal` because we have `bicoloring => !has_diagonal`

amontoison:
> I suggest replacing `AdjacencyGraph` with:
> 
> ```julia
> struct AdjacencyGraph{T<:Integer,bicoloring}
>     S::SparsityPatternCSC{T}
>     edge_to_index::Vector{T}
>     original_size::Tuple{Int,Int}
> end
> ```

Ok for me 👍

> We no longer need `has_diagonal` because we have `bicoloring => !has_diagonal`

Yes but we don't have `!has_diagonal => bicoloring`. Is it fine for you that the user can't specify anymore `has_diagonal = false` in a non-bicoloring case?

I think adding a field `bicoloring::Bool` can handle more corner cases.

gdalle:
> Yes but we don't have !has_diagonal => bicoloring. Is it fine for you that the user can't specify anymore has_diagonal = false in a non-bicoloring case?

`has_diagonal` was never part of our public API to begin with, so the user never gets to specify it. I think the cases where someone wants to color a Hessian whose entire diagonal is zero are sufficiently rare for us to ignore them.
In addition, they would lead to type instabilities because they require a check on the values of the matrix before creating the `AdjacencyGraph` object.

test/suitesparse.jl
```diff
@@ -143,6 +143,7 @@ what_table_41_42 = CSV.read(
         @test nb_edges(ag) == row[:E]
         @test maximum_degree(ag) == row[:Δ]
         @test minimum_degree(ag) == row[:δ]
+        bicoloring = false
```
amontoison:
```suggestion
```

src/postprocessing.jl
```diff
@@ -98,7 +99,8 @@ function postprocess_star_coloring!(
 
     # Process the trivial stars (if any)
     if nb_trivial_stars > 0
-        nb_unknown_hubs = nb_trivial_stars
+        occurrences .= 0
```
gdalle:
```suggestion
        fill!(occurrences, 0)
```
is slightly more efficient

test/small.jl
```diff
@@ -181,18 +181,20 @@ end;
             ),
         )
 
-        test_bicoloring_decompression(
-            A,
-            problem,
-            GreedyColoringAlgorithm{:direct}(order; postprocessing=true);
-            test_fast=true,
-        )
+        @testset "postprocessing_minimizes = $target" for target in (:all_colors, :row_colors, :column_colors)
```
gdalle: This should be added to the random tests too

src/postprocessing.jl
```diff
@@ -0,0 +1,550 @@
+## Postprocessing
+
+function postprocess!(
+    color::AbstractVector{<:Integer},
+    star_or_tree_set::Union{StarSet,TreeSet},
```
gdalle: The splitting of the code into 4 functions makes it difficult to review and maintain. I think we can get away with only two, separating `StarSet` vs `TreeSet` but keeping a unified logic for symmetric and bidirectional. We can fill a `column_color_used` and `row_color_used` in both cases, and then recover `color_used` with a `max` operation. That would make things much clearer

gdalle (changes requested):
Aside from my remarks above, do you have any ideas for test cases where our heuristics actually produce the result we want?

src/postprocessing.jl
```diff
@@ -0,0 +1,550 @@
...
+    g::AdjacencyGraph,
+    offsets::AbstractVector{<:Integer},
+    postprocessing_minimizes::Symbol,
+)
+    # flag which colors are actually used during decompression
+    nb_colors = maximum(color)
+    color_used = zeros(Bool, nb_colors)
+
+    # nonzero diagonal coefficients force the use of their respective color (there can be no neutral colors if the diagonal is fully nonzero)
+    if has_diagonal(g)
```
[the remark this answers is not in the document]
amontoison: Yep I agree but it is not directly related to this PR.
I opened an issue for that.

src/postprocessing.jl
amontoison: It is what I asked you Guillaume during our last meeting and you said yes and even suggested to split again for bicoloring.

amontoison: I will do my best to fuse the functions such we can keep only 2.

gdalle: > It is what I asked you Guillaume during our last meeting and you said yes and even suggested to split again for bicoloring.

I thought we would only get one split, now two, sorry for the misunderstanding

test/small.jl
amontoison: Done in this commit: https://github.com/gdalle/SparseMatrixColorings.jl/pull/267/commits/5f40c3dc5eb60a69bef0e3698ac49df1d1c2ab1f

amontoison:
@gdalle It is finally ready for a review!!

amontoison:
@gdalle I was able to run the code in the two lines not covered by the unit tests with a random order and the following matrices:
```julia
6×12 SparseMatrixCSC{Float64, Int64} with 11 stored entries:
  ⋅   1.0   ⋅    ⋅    ⋅   1.0   ⋅    ⋅   1.0   ⋅    ⋅    ⋅ 
  ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅   1.0   ⋅    ⋅    ⋅    ⋅ 
  ⋅   1.0  1.0   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅ 
  ⋅    ⋅   1.0   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅ 
  ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅   1.0   ⋅    ⋅   1.0
  ⋅    ⋅    ⋅   1.0   ⋅    ⋅   1.0   ⋅    ⋅    ⋅    ⋅    ⋅ 

6×12 SparseMatrixCSC{Float64, Int64} with 11 stored entries:
  ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅ 
  ⋅    ⋅    ⋅    ⋅    ⋅    ⋅   1.0   ⋅    ⋅   1.0  1.0   ⋅ 
  ⋅    ⋅    ⋅    ⋅   1.0   ⋅    ⋅    ⋅   1.0   ⋅    ⋅    ⋅ 
  ⋅    ⋅   1.0   ⋅    ⋅    ⋅    ⋅    ⋅   1.0   ⋅    ⋅    ⋅ 
  ⋅    ⋅    ⋅   1.0   ⋅   1.0   ⋅   1.0   ⋅    ⋅    ⋅    ⋅ 
  ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅   1.0   ⋅    ⋅    ⋅    ⋅

3×6 SparseMatrixCSC{Float64, Int64} with 6 stored entries:
 1.0   ⋅   1.0  1.0   ⋅    ⋅ 
  ⋅    ⋅    ⋅    ⋅   1.0   ⋅ 
  ⋅    ⋅    ⋅   1.0   ⋅   1.0

3×6 SparseMatrixCSC{Float64, Int64} with 5 stored entries:
 1.0   ⋅    ⋅    ⋅    ⋅    ⋅ 
  ⋅   1.0  1.0   ⋅    ⋅   1.0
  ⋅   1.0   ⋅    ⋅    ⋅    ⋅ 

3×6 SparseMatrixCSC{Float64, Int64} with 7 stored entries:
  ⋅    ⋅    ⋅    ⋅   1.0   ⋅ 
  ⋅    ⋅   1.0   ⋅    ⋅   1.0
 1.0  1.0  1.0   ⋅   1.0   ⋅ 

3×5 SparseMatrixCSC{Float64, Int64} with 5 stored entries:
  ⋅   1.0  1.0  1.0   ⋅ 
 1.0   ⋅    ⋅    ⋅    ⋅ 
  ⋅    ⋅   1.0   ⋅    ⋅ 

3×4 SparseMatrixCSC{Float64, Int64} with 5 stored entries:
  ⋅   1.0  1.0   ⋅ 
  ⋅    ⋅    ⋅   1.0
 1.0   ⋅   1.0   ⋅

3×4 SparseMatrixCSC{Float64, Int64} with 6 stored entries:
  ⋅    ⋅    ⋅   1.0
 1.0  1.0   ⋅    ⋅ 
 1.0   ⋅   1.0  1.0

3×4 SparseMatrixCSC{Float64, Int64} with 4 stored entries:
  ⋅    ⋅   1.0  1.0
  ⋅    ⋅   1.0   ⋅ 
 1.0   ⋅    ⋅    ⋅ 

3×4 SparseMatrixCSC{Float64, Int64} with 6 stored entries:
 1.0  1.0   ⋅   1.0
 1.0   ⋅    ⋅    ⋅ 
  ⋅   1.0  1.0   ⋅ 

3×3 SparseMatrixCSC{Float64, Int64} with 4 stored entries:
  ⋅   1.0   ⋅ 
 1.0   ⋅   1.0
  ⋅    ⋅   1.0

3×3 SparseMatrixCSC{Float64, Int64} with 4 stored entries:
 1.0   ⋅   1.0
 1.0   ⋅    ⋅ 
  ⋅   1.0   ⋅ 

3×3 SparseMatrixCSC{Float64, Int64} with 4 stored entries:
 1.0  1.0   ⋅ 
  ⋅    ⋅   1.0
  ⋅   1.0   ⋅
 ```
 I will add the 3x3 matrix in the tests and we should be good.

gdalle:
Thanks!
I'm still on vacation but I'll review when I get back to work

gdalle:
Thanks for running the tests! If I understand correctly, the takeaways are:

1. The new postprocessing reduces the number of colors for a larger set of instances
2. It is unclear whether the `postprocessing_minimizes` argument actually changes anything
3. The order has a big impact, which to me suggests that we should try using it not just when we initially assign colors, but also when we go back through the list of vertices to mark them as essential or not (e.g. [here](https://github.com/gdalle/SparseMatrixColorings.jl/blob/87c8016a57048a221e404fdf75f5e27621242a99/src/postprocessing.jl#L114))

gdalle:
> What is important is the order of treatment of the trivial trees or stars (last step of post-processing).

Sorry, yes, replace the line I pointed to with e.g. [these ones](https://github.com/gdalle/SparseMatrixColorings.jl/blob/87c8016a57048a221e404fdf75f5e27621242a99/src/postprocessing.jl#L138-L139). Perhaps there we could permute stars or trees according to the same order used for coloring, or its reverse order?

amontoison:
> > What is important is the order of treatment of the trivial trees or stars (last step of post-processing).
> 
> Sorry, yes, replace the line I pointed to with e.g. [these ones](https://github.com/gdalle/SparseMatrixColorings.jl/blob/87c8016a57048a221e404fdf75f5e27621242a99/src/postprocessing.jl#L138-L139). Perhaps there we could permute stars or trees according to the same order used for coloring, or its reverse order?

For the acyclic coloring, we store ane iterate over the list of trees, so I don't have any idea of how we could use the ordering of the vertices for that.

Foe the star coloring, we iterate over the stars multiple times. We iterate over the star of each vertex and not the stars directly (we can't because we only encode the hub and not the spokes in our implementation of the stars).
But what we want in pratice is to iterate over the stars and I don't know how we can permute them based on the ordering of the vertices.

It is an opened question and the treatment of the trivial stars / trees is a standalone new graph problem from what I understand.

We should not forget to explain it in our paper when we will get the review.

amontoison:
@gdalle We should merge this PR and move forward.
I am confident about the content.

gdalle:
If you're confident I trust your opinion, I agree that it's been cooking for too long

gdalle (approved):
I approve this PR trusting Alexis' opinion (didn't take another close look, 2026 was kind of a blur)

PR #288, opened 2025-12-31
Enhance decompress! for bicoloring

amontoison:


amontoison:
I forgot to run the benchmarks on this PR.
But I am not sure that we run it for matrices that are not `SparseMatrixCSC`.
I expect a big gain in terms of storage.

Issue #292, opened 2026-01-04
SMC.jl is not working with CUDA.jl 5.9.6


amontoison:
I highly suspect that the culprit is this PR: https://github.com/JuliaGPU/CUDA.jl/pull/2942

Confirmed with the dummy PR here: #291

gdalle:
Open an issue in CUDA ?

amontoison:
Yes, but I need to prepare a MWE.
I wanted to do something else than GPU stuff tonight.

PR #293, opened 2026-01-04
Add a compat entry for CUDA.jl in test/Project.toml

amontoison:
Temporary solution for #292


amontoison:
I fixed it such that I don't break it with other modifications in `SMC.jl`.
I think it is still relevant to keep the tests working (if it doesn't hold back the ecosystem).

PR #274, opened 2025-11-01
Support sparse matrices from rocSPARSE

amontoison:
close https://github.com/gdalle/SparseMatrixColorings.jl/issues/254
Very similar to https://github.com/gdalle/SparseMatrixColorings.jl/pull/253.
Supersede #255

PR #294, opened 2026-01-04
Add an option decompression_uplo for symmetric results

amontoison:
Add an argument `decompression_uplo` for `StarSetColoringResult` and `TreeSetColoringResult` such that we can specialize the decompression for bicoloring.
We can always use `decompression_uplo = :L` for the bicoloring.
For the bicoloring, `decompress!` only use the specialized version for `SparseMatrixCSC` since #288.

Everything is internal so it is not breaking.
Replace #289.

I will wait a review of Guillaume before that I merge this PR and future modifications.

### DIFF
--- a/src/decompression.jl
+++ b/src/decompression.jl
@@ -448,6 +448,7 @@ end
 function decompress!(
     A::AbstractMatrix, B::AbstractMatrix, result::StarSetColoringResult, uplo::Symbol=:F
 )
+    @assert result.decompression_uplo == :F
     (; ag, compressed_indices) = result
     (; S) = ag
     check_compatible_pattern(A, ag, uplo)
@@ -472,6 +473,7 @@ function decompress_single_color!(
     result::StarSetColoringResult,
     uplo::Symbol=:F,
 )
+    @assert result.decompression_uplo == :F
     (; ag, compressed_indices, group) = result
     (; S) = ag
     check_compatible_pattern(A, ag, uplo)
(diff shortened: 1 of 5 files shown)


amontoison:
Yes, it should make the `uplo` argument useless (if we specialize it in the result for `:L` or `:U`).
But for bicoloring, we don't specify the decompression uplo, it is always `:L` by design, we don't have issue here.

The one in the result always takes precedence.
Note that it is not exposed to the users for star and acyclic coloring with main function `coloring`.

gdalle:
I took a second look and the semantics of this are a bit confusing to me. From what I understand:

- If you create the result with `decompression_uplo = :L`, you allow yourself to optimise the construction because you _know_ that it will be used only inside bicoloring, and only with `uplo = decompression_uplo = :L`
- But if you create the result with `decompression_uplo = :F`, then you must still account for the possibility that `uplo != decompression_uplo`?

src/decompression.jl
```diff
@@ -509,11 +511,12 @@ function decompress!(
     (; S) = ag
     nzA = nonzeros(A)
     check_compatible_pattern(A, ag, uplo)
-    if uplo == :F
+    if result.decompression_uplo == uplo
```
gdalle: Why is this case not considered for the `TreeSetColoringResult` below?

gdalle (changes requested):

src/decompression.jl
amontoison: Because we only gain on the storage when the decompression is by substitution.
The code for the decompression itself is not impacted.

However, I forgot to add an assertion.
I fixed it in my last commit such that:
```julia
function decompress!(
    A::SparseMatrixCSC{R},
    B::AbstractMatrix{R},
    result::TreeSetColoringResult,
    uplo::Symbol=:F,
) where {R<:Real}
    check_compatible_pattern(A, result.ag, uplo)
    @assert result.decompression_uplo == uplo || result.decompression_uplo == :F
    decompress_csc!(nonzeros(A), A.colptr, B, result, uplo)
    return A
end
```

amontoison:
I propose to merge first the support of AMD GPUs (#274).
A minor modification is needed for the extensions and I would to do it for both backends in the same PR.

PR #296, opened 2026-01-09
Document fields of TreeSet

blegat:
I needed to understand these from the implementation for the purpose of https://github.com/blegat/ArrayDiff.jl/pull/19 so I thought I might as well document them now that I understand what they represent :)

### DIFF
--- a/src/coloring.jl
+++ b/src/coloring.jl
@@ -423,7 +423,6 @@ $TYPEDFIELDS
 """
 struct TreeSet{T}
     """
-    contains the reverse breadth first (BFS) traversal order for each tree in the forest.
     For a tree index `1 <= k <= nt`, the list
     `list = reverse_bfs_order[tree_edge_indices[k]:(tree_edge_indices[k+1]-1)]` is a list of edges
     `list[i] = (leaf, inner)` of the `k`th tree such that `leaf` is a leaf of the tree containing


amontoison:
It needs a continuous view because JuMP is doing a coloring for the Hessian of the objective and then for the Hessian of each constraints.
They are not doing like `ADNLPModels` where we do directly a coloring of the Hessian of the Lagrangian.
After the computation of the nonzeros of each compressed Hessian, they need to do the decompression in the related part of the vector `V` to recover the sparse Hessian.

@blegat I am right?

src/coloring.jl
```diff
@@ -566,7 +576,6 @@ function TreeSet(
     # Number of edges treated
     num_edges_treated = zero(T)
 
-    # reverse_bfs_orders contains the reverse breadth first (BFS) traversal order for each tree in the forest
```
amontoison: We should keep this comment.
What do think of:
```
# reverse_bfs_orders will contain the reverse breadth first (BFS) traversal order for each tree in the forest concatenated
```

blegat: I added a new comment clarifying the confusion with the root of the `Forest` union-find datastructure that we thought we be useful during our discussion on the phone

amontoison (approved):

PR #298, opened 2026-01-19
Accept AbstractSparseMatrixCSC for decompression

blegat:
Here is a suggestion for https://github.com/gdalle/SparseMatrixColorings.jl/pull/296#issuecomment-3729410871
I can write tests if the change sounds reasonable

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -46,7 +46,6 @@ compress
 decompress
 decompress!
 decompress_single_color!
-decompress_csc!
 ```
 
 ## Orders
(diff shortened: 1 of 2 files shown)


gdalle:
I think I'd rather add an internal indirection step like
```julia
function decompress!(A::SparseMatrixCSC, ...)
    return decompress_csc!(A.nzval, A.rowval, A.colptr, A.m, A.n, ...)
end
```
than rely on an incompletely specified `AbstractSparseMatrixCSC` API. You can then use `decompress_csc!` in ArrayDiff, until we figure out something cleaner?

PR #299, opened 2026-01-19
Fix respectful_similar with SparsityPatternCSC

blegat:
Otherwise, `decompress` returns a dense matrix when a `SparsityPatternCSC` is given as argument to `coloring`, see below:
```julia
julia> S = sparse([
           0 0 1 1 0 1
           1 0 0 0 1 0
           0 1 0 0 1 0
           0 1 1 0 0 0
       ])
4×6 SparseMatrixCSC{Int64, Int64} with 9 stored entries:
 ⋅  ⋅  1  1  ⋅  1
 1  ⋅  ⋅  ⋅  1  ⋅
 ⋅  1  ⋅  ⋅  1  ⋅
 ⋅  1  1  ⋅  ⋅  ⋅

julia> P = SparseMatrixColorings.SparsityPatternCSC(S)
4×6 SparseMatrixColorings.SparsityPatternCSC{Int64}:
 false  false   true   true  false   true
  true  false  false  false   true  false
 false   true  false  false   true  false
 false   true   true  false  false  false

julia> problem = ColoringProblem()
ColoringProblem{:nonsymmetric, :column}()

julia> algo = GreedyColoringAlgorithm()
GreedyColoringAlgorithm{:direct, 1, Tuple{NaturalOrder}}((NaturalOrder(),), false)

julia> result = coloring(P, problem, algo); # We give P, not S

julia> B = compress(S, result)
4×3 Matrix{Int64}:
 1  1  1
 1  1  0
 1  1  0
 1  1  0

julia> C = decompress(B, result)
4×6 Matrix{Int64}:
 0  0  1  1  0  1
 1  0  0  0  1  0
 0  1  0  0  1  0
 0  1  1  0  0  0
```
This affects https://github.com/blegat/ArrayDiff.jl/pull/23 because I just create a `SparsityPatternCSC` with the detected sparsity pattern and use `coloring` with this as input.

I can write tests if the change sounds reasonable

### DIFF
--- a/test/structured.jl
+++ b/test/structured.jl
@@ -2,6 +2,7 @@ using ArrayInterface: ArrayInterface
 using BandedMatrices: BandedMatrix, brand
 using BlockBandedMatrices: BandedBlockBandedMatrix, BlockBandedMatrix
 using LinearAlgebra
+using SparseArrays
 using SparseMatrixColorings
 using Test
 
@@ -58,8 +59,8 @@ end;
 end;
 
 # See https://github.com/gdalle/SparseMatrixColorings.jl/pull/299
-@testset "SparsityPatternCSC" begin
-    S = sparse([
+@testset "SparsityPatternCSC $T" for T in [Int, Float32]
+    S = sparse(T[
         0 0 1 1 0 1
         1 0 0 0 1 0
         0 1 0 0 1 0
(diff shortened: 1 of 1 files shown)


gdalle:
It should probably also return a SparseMatrixCSC, that way values are modifiable with a surface API

amontoison (approved):

PR #139, opened 2024-10-03
Optimized implementation for structured matrices

gdalle:
Fixes #64 and fixes #65 by optimizing structured coloring & decompression for these matrix types:

| Package | Type | Optimal coloring | Optimal decompression |
|---|---|---|---|
| LinearAlgebra | `Diagonal` | ✅ | ✅ | 
| LinearAlgebra | `Bidiagonal` | ✅ | ✅ | 
| LinearAlgebra | `Tridiagonal` | ✅ | ✅ | 
| BandedMatrices | `BandedMatrix` | ✅ | no need |
| BlockBandedMatrices | `BlockBandedMatrix` | ❓ | no need |
| BlockBandedMatrices | `BandedBlockBandedMatrix` | ✅ |  |

For `BandedMatrix` and `BlockBandedMatrix`, the standard decompression based on `SparseMatrixCSC` will work fine because their storage is column-major ([ref1](https://julialinearalgebra.github.io/BandedMatrices.jl/stable/#Implementation), [ref2](https://julialinearalgebra.github.io/BlockBandedMatrices.jl/stable/#Implementation)). We only need a specific overload for `BandedBlockBandedMatrix`, which is a `BlockMatrix` of `BandedMatrix`-like blocks (so we need to decompress block by block).

**General philosophy**

Separate coloring from decompression:

- For coloring, define a new `StructuredColoringAlgorithm` that uses optimal coloring formulas (note that for block-banded the coloring given here is not actually optimal if the blocks themselves are sparse).
- For decompression, leverage the usual result type (`ColumnColoringResult` / `RowColoringResult`), which means that even greedy colorings on structured matrices benefit from optimized decompression.

**Remaning issues**

For decompression of BlockBandedMatrices, I still need to figure out https://github.com/JuliaArrays/BlockArrays.jl/issues/418. But it already works because the fallback algorithm relies on the nonzeros of the sparse encoding, so it does not try to access illegal indices outside of the bands.

**Implementation details**

This is a clean rewrite of some code that was already present in ArrayInterface.jl and FiniteDiff.jl:

https://github.com/JuliaArrays/ArrayInterface.jl/blob/2e2b81bfb53e9e9fc41f26b927fb980afcbfa9f5/src/ArrayInterface.jl#L372-L384

https://github.com/JuliaArrays/ArrayInterface.jl/blob/2e2b81bfb53e9e9fc41f26b927fb980afcbfa9f5/ext/ArrayInterfaceBandedMatricesExt.jl#L68-L72

[...]

PR #319, opened 2026-04-29
Optimized implementation for structured matrices V2

ErikQQY:
Basically #139 with new implementations

### DIFF
--- a/ext/SparseMatrixColoringsCUDAExt.jl
+++ b/ext/SparseMatrixColoringsCUDAExt.jl
@@ -1,9 +1,30 @@
 module SparseMatrixColoringsCUDAExt
-
+using LinearAlgebra
 import SparseMatrixColorings as SMC
 using SparseArrays: SparseMatrixCSC, rowvals, nnz, nzrange
-using CUDA: CuVector, CuMatrix
+using CUDA: CuArray, CuVector, CuMatrix
 using cuSPARSE: AbstractCuSparseMatrix, CuSparseMatrixCSC, CuSparseMatrixCSR
+
+## Basic support for GPU sparsity pattern stuff
+
+SMC.SparsityPatternCSC(A::CuSparseMatrixCSC) = SMC.SparsityPatternCSC(first(A.dims), last(A.dims), A.colPtr, A.rowVal)
+
+for R in (:Diagonal, :Bidiagonal, :Tridiagonal)
+    @eval function SMC.BipartiteGraph(A::$R{T, <:CuArray}; symmetric_pattern::Bool=false) where {T}
+        return SMC.BipartiteGraph(CuSparseMatrixCSC(A); symmetric_pattern)
+    end
+end
+
+function SMC.BipartiteGraph(A::CuSparseMatrixCSC; symmetric_pattern::Bool=false)
+    S2 = SMC.SparsityPatternCSC(A)
+    if symmetric_pattern
+        checksquare(A)  # proxy for checking full symmetry
+        S1 = S2
+    else
+        S1 = transpose(S2)  # rows to columns
+    end
+    return SMC.BipartiteGraph(S1, S2)
+end
 
 ## CSC Result
 
--- a/ext/SparseMatrixColoringsCUDAExt.jl
+++ b/ext/SparseMatrixColoringsCUDAExt.jl
@@ -7,10 +7,14 @@ using cuSPARSE: AbstractCuSparseMatrix, CuSparseMatrixCSC, CuSparseMatrixCSR
 
 ## Basic support for GPU sparsity pattern stuff
 
-SMC.SparsityPatternCSC(A::CuSparseMatrixCSC) = SMC.SparsityPatternCSC(first(A.dims), last(A.dims), A.colPtr, A.rowVal)
+function SMC.SparsityPatternCSC(A::CuSparseMatrixCSC)
+    SMC.SparsityPatternCSC(first(A.dims), last(A.dims), A.colPtr, A.rowVal)
+end
 
 for R in (:Diagonal, :Bidiagonal, :Tridiagonal)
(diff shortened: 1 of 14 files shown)


ext/SparseMatrixColoringsCUDAExt.jl
```diff
@@ -1,10 +1,31 @@
...
+    S2 = SMC.SparsityPatternCSC(A)
+    if symmetric_pattern
+        checksquare(A)  # proxy for checking full symmetry
+        S1 = S2
+    else
+        S1 = transpose(S2)  # rows to columns
+    end
+    return SMC.BipartiteGraph(S1, S2)
+end
+
```
ErikQQY, line 32: @gdalle For now I just want to make sure the structured coloring algorithms working for Diagonal, Bidiagonal and Tridiagonal matrices, but it turns out we still need to make `transpose` to be GPU-compatiable to allow colorings and decompression. Do these implementations look resonable to you, if do, I will continue on this direction.

ext/SparseMatrixColoringsColorsExt.jl
```diff
@@ -271,8 +271,9 @@ function show_colors!(
     A_ccolor_indices = mod1.(column_colors(res), length(colorscheme))
     A_rcolor_indices = mod1.(row_shift .+ row_colors(res), length(colorscheme))
     B_ccolor_indices = mod1.(1:maximum(column_colors(res)), length(colorscheme))
-    B_rcolor_indices =
-        mod1.((row_shift + 1):(row_shift + maximum(row_colors(res))), length(colorscheme))
+    B_rcolor_indices = mod1.(
+        (row_shift + 1):(row_shift + maximum(row_colors(res))), length(colorscheme)
+    )
```
gdalle, line 276: If possible, it's better to avoid formatting changes that only add noise

src/graph.jl
```diff
@@ -17,8 +17,8 @@ Copied from `SparseMatrixCSC`:
 struct SparsityPatternCSC{Ti<:Integer} <: AbstractMatrix{Bool}
     m::Int
     n::Int
-    colptr::Vector{Ti}
-    rowval::Vector{Ti}
+    colptr::AbstractVector{Ti}
+    rowval::AbstractVector{Ti}
```
gdalle, line 21: This is a really bad idea because it will introduce type instability throughout the coloring and decompression algorithms. Can you explain why you deem it necessary?
I hope the tests catch it, if not we need more type stability tests with JET

ext/SparseMatrixColoringsJuMPExt.jl
```diff
@@ -29,7 +29,7 @@ function optimal_distance2_coloring(
     model = Model(optimizer)
     silent && set_silent(model)
     # one variable per vertex to color, removing some renumbering symmetries
-    @variable(model, 1 <= color[i=1:n] <= i, Int)
+    @variable(model, 1 <= color[i = 1:n] <= i, Int)
```
gdalle, line 32: Same comment about formatting noise

test/utils.jl
```diff
@@ -287,9 +288,8 @@ function test_structured_coloring_decompression(A::AbstractMatrix)
     @test D == A
     @test nameof(typeof(D)) == nameof(typeof(A))
     @test structurally_orthogonal_columns(A, color)
-    if VERSION >= v"1.10" || A isa Union{Diagonal,Bidiagonal,Tridiagonal}
-        # banded matrices not supported by ArrayInterface on Julia 1.6
-        # @test color == ArrayInterface.matrix_colors(A)  # TODO: uncomment
+    if algo isa StructuredColoringAlgorithm
```
gdalle, line 291: We should also restrict this to the matrix types that ArrayInterface supports

test/utils.jl
```diff
@@ -287,9 +288,8 @@ function test_structured_coloring_decompression(A::AbstractMatrix)
...
+        @test color == ArrayInterface.matrix_colors(A)
```
gdalle, line 292: We may also want to test that decompression returns a matrix of the correct type

src/structured.jl
```diff
@@ -0,0 +1,181 @@
...
+    return A
+end
+
+function decompress!(A::Diagonal, B::AbstractMatrix, result::RowColoringResult)
+    color = row_colors(result)
+    for i in axes(A, 1)
+        A[i, i] = B[color[i], i]
+    end
+    return A
+end
```
gdalle, line 75: This will not work for a `Diagonal` matrix stored on GPU. My idea was that we could vectorize index accesses like we do for the CSC/CSR formats on GPU, and then do something like
```julia
diag(A) .= getindex.(B, color, 1:n)
```

ext/SparseMatrixColoringsCUDAExt.jl
```diff
@@ -1,10 +1,35 @@
...
+    SMC.SparsityPatternCSC(first(A.dims), last(A.dims), A.colPtr, A.rowVal)
+end
+
+for R in (:Diagonal, :Bidiagonal, :Tridiagonal)
+    @eval function SMC.BipartiteGraph(
+        A::$R{T,<:CuArray}; symmetric_pattern::Bool=false
+    ) where {T}
+        return SMC.BipartiteGraph(CuSparseMatrixCSC(A); symmetric_pattern)
+    end
+end
```
gdalle, line 20: Why is this necessary? The coloring will take place on the CPU anyway so I don't see why we would need a GPU-based graph structure

ext/SparseMatrixColoringsCUDAExt.jl
gdalle: See above, I think I need additional context to understand what you did here

gdalle (changes requested):
Thanks for getting this started!

PR #320, opened 2026-05-26
Add pre-commit workflow for formatting

gdalle:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
+- repo: https://github.com/JuliaEditorSupport/JuliaFormatter.jl
+  rev: 3b196afa4a8ed92ba816edec6d122145e311ae35 # This PR.
+  hooks:
+  - id: "jlfmt"
+    args: ["--jlfmt-path=~/.julia/bin/jlfmt"]
```
penelopeysm: Hmmmm - looking at this I think I could make it easier by hardcoding this path as a backup in case plain `jlfmt` isn't found on PATH. For like 99% of use cases it will be here anyway... Then one could also check JULIA_DEPOT_PATH, etc

gdalle: Yeah that's probably what I would suggest. Is it possible to set this up interpolating `JULIA_DEPOT_PATH` instead of hardcoding `.julia`?

penelopeysm: Yup! I'll edit the script to search in the following order:

- manually specified path to executable
- jlfmt on PATH
- check if `$JULIA_DEPOT_PATH` is set first and search in there
- check `$HOME/.julia/bin` (which Julia uses as the default if the envvar isn't set - I just checked https://github.com/JuliaLang/julia/blob/32a8439612b3794c58b0bfea1047876e7c0ee6b3/base/initdefs.jl#L169-L174)

.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
+- repo: https://github.com/JuliaEditorSupport/JuliaFormatter.jl
+  rev: 3b196afa4a8ed92ba816edec6d122145e311ae35 # This PR.
```
penelopeysm:
```suggestion
  rev: 0935890389c9fbc45f4ee150064c0abe5340b6a9 # v2.4.0
```

.pre-commit-config.yaml
penelopeysm:
```suggestion
  - id: "jlfmt"
```

penelopeysm (commented):
Fingers crossed?

